PR #321, opened 2021-01-15
Draft: Handle generic type hinting in PHPDoc

dsavina:
This allows declaring iterable types in PHPDoc as `IterableType<IteratedType>` instead of having to do something like `IterableType|IteratedType[]`.

### DIFF
--- a/src/Mappers/Root/IteratorTypeMapper.php
+++ b/src/Mappers/Root/IteratorTypeMapper.php
@@ -15,7 +15,7 @@ use Iterator;
 use IteratorAggregate;
 use phpDocumentor\Reflection\DocBlock;
 use phpDocumentor\Reflection\Type;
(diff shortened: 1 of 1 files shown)


dsavina:
@moufmouf, I would only need to know where/how to implement the test case for this feature to remove the `Draft` tag on this request (I have tested it in a project, though).

oojacoboo:
@dsavina I'm not sure, but I think the `tests/Types/InputTypeTest` class would probably be a good spot for this.  You should get a better idea as you're writing the test.

Lappihuan:
@dsavina `tests/Mappers/Root/IteratorTypeMapperTest.php` has a bunch of tests regarding this mapper

oojacoboo:
Generic typed iterable support was provided in https://github.com/thecodingmachine/graphqlite/pull/468

This has been released in version 6.0.

PR #329, opened 2021-02-16
Update CannotMapTypeException.php

gulien:
^ better error message for input type.


Lappihuan:
by now the suggestion can't be hardcoded anymore since Factories are not the only way to create input types.
this should probably be closed for now.

PR #358, opened 2021-08-05
Docs: Add information about class mapper

katakumby:
I think it would be great to include information about setClassNameMapper property into docs as well

### DIFF
--- a/website/versioned_docs/version-4.1/other_frameworks.md
+++ b/website/versioned_docs/version-4.1/other_frameworks.md
@@ -54,6 +54,8 @@ The `SchemaFactory` class also comes with a number of methods that you can use t
 $factory->setAuthenticationService(new VoidAuthenticationService());
 // Configure an authorization service (to resolve the @Right annotations).
 $factory->setAuthorizationService(new VoidAuthorizationService());
+//Add class mapper to map a class names to one or many possible file names according to PSR-0 or PSR-4
+$factory->setClassNameMapper(new ClassNameMapper());
 // Change the naming convention of GraphQL types globally.
 $factory->setNamingStrategy(new NamingStrategy());
 // Add a custom type mapper.


oojacoboo:
@katakumby Thanks for the PR.  I think under most circumstances this isn't a necessity at all.  Your composer autoload should take care of this.  Do you have a specific use case here?  Or, is this more about documenting the public API?  I'm not entirely opposed to this, but it does not need to be made so prominent in the config docs, and should be prefaced with information about Composer, the preferred way of handling the mapping.  We don't really want people abusing this.

katakumby:
This is more about documenting public API and you are right it don't have to be so prominent, in my case i needed to register custom directory because of customized non framework environment, when using that method i was able to properly integrate https://github.com/ThingEngineer/PHP-MySQLi-Database-Class existing models to support graphqlite in separated environment's with shared models, im doing it that way to have as less as possible dependencies across environments/apps.

website/versioned_docs/version-4.1/other_frameworks.md
```diff
@@ -54,6 +54,8 @@ The `SchemaFactory` class also comes with a number of methods that you can use t
 $factory->setAuthenticationService(new VoidAuthenticationService());
 // Configure an authorization service (to resolve the @Right annotations).
 $factory->setAuthorizationService(new VoidAuthorizationService());
+//Add class mapper to map a class names to one or many possible file names according to PSR-0 or PSR-4
```
oojacoboo, line 57: Please follow the comment code-style, space after `//`

website/versioned_docs/version-4.1/other_frameworks.md
```diff
@@ -54,6 +54,8 @@ The `SchemaFactory` class also comes with a number of methods that you can use t
...
+$factory->setClassNameMapper(new ClassNameMapper());
```
oojacoboo, line 58: Please move this further down the calls, ie. priority.

oojacoboo (changes requested):

oojacoboo:
The docs have been updated quite a bit.  Please review for merging.

PR #362, opened 2021-08-14
Fix CI

TamasSzigeti:
* Run tests for all php versions, lowest and highest deps
* Do not run on branch push, only PR scope
* Add cache for phpstan
* Fix dependency issues
* Fix stan errors
* Only run phpcs for lowest php version
* Fix php 7.2 compatibility

I didn't fix the graphqlite-universal-service-provider job, probably need to address it in the other repo


TamasSzigeti:
@moufmouf @devmaslov @oojacoboo Please kindly approve workflow run and review. 
More is coming after this

.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea
```
oojacoboo, line 1: While having this in the .gitignore isn't particularly bad, developer specific ignores should really be in your global gitignores.  Please see the following article for details:  https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

composer.json
```diff
@@ -11,38 +12,38 @@
...
-        "thecodingmachine/class-explorer": "^1.1.0",
-        "psr/simple-cache": "^1",
+        "composer/package-versions-deprecated": "^1.8",
         "phpdocumentor/reflection-docblock": "^4.3 || ^5.0",
-        "phpdocumentor/type-resolver": "^1.0.1",
-        "psr/http-message": "^1",
-        "webmozart/assert": "^1.4",
-        "symfony/cache": "^4.3 | ^5",
-        "thecodingmachine/cache-utils": "^1",
-        "ocramius/package-versions": "^1.4 || ^2",
```
oojacoboo, line 26: Was this no longer needed?

oojacoboo:
@TamasSzigeti thanks for taking the initiative with this BTW.

composer.json
TamasSzigeti: I replaced it with its fork to get rid of errors. I already did this in graphqlite-bundle

.gitignore
TamasSzigeti: I disagree, some projects prefer sharing ide settings, others do not, not really a global setting. But I can drop it if you wish

TamasSzigeti: Done

oojacoboo:
@TamasSzigeti looks like there is just one more dep conflict.  Thanks for working on this.

TamasSzigeti:
@oojacoboo I wrote above that the `universal` workflow will need fixing in that repo. Or am I wrong?

oojacoboo:
@TamasSzigeti I see now.  I've expressed my desires to have these external child dependencies removed from this lib.  IMO, this lib does not need to be hindered by these other "wrapper" libs.  @moufmouf already mentioned that the `graphqlite-universal-service-provider` repo is basically not used and virtually abandoned.

If you wouldn't mind removing that check, that'd be great.  Then we can get all these checks passing.

I'm in favor of also removing the symfony checks as well, for the reasons stated above.  But, we can leave them for now.

TamasSzigeti:
@oojacoboo Done. For now skipped, in case we want to revisit.
Any way I can get permanent approval for workflows? I plan to contribute more and more (:

oojacoboo:
@TamasSzigeti I believe @moufmouf will have to update the settings or change my role here.  I'm not seeing the settings for this available to me.

TamasSzigeti:
@oojacoboo We should also switch to squash&merge strategy, now all my messy commits are in master (:

oojacoboo:
@TamasSzigeti yea, that's my bad.  I agree, we'll do that going forward.  Thanks for the heads up.

PR #399, opened 2021-10-26
Added TARGET_PARAMETER for PHP8 Attributes

### LINKED ISSUE #354: Allow Field (and related) annotation on constructor properties
It would be really nice to do this:

```php
<?php

#[Type]
class AType {
  public function __construct(#[Field] public string $field) {}
}
```

especially for really simple types where you don't need any behavior on getters.

Lappihuan:
As discussed in the [RFC of constructor property promotion](https://wiki.php.net/rfc/constructor_promotion#attributes) this would allow the #[Field] attribute to be used in function parameters but would automatically be picked up by graphqlite because it would find a Field Attribute with ReflectionProperty.

Would resolve #354

### DIFF
--- a/src/Annotations/Field.php
+++ b/src/Annotations/Field.php
@@ -18,7 +18,7 @@ use Attribute;
  *   @Attribute("inputType", type = "string"),
  * })
  */
-#[Attribute(Attribute::TARGET_PROPERTY | Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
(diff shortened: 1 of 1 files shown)


oojacoboo:
@Lappihuan Feel free to commit to the branch you have, allowing Github to run the tests here in this PR.

Lappihuan:
i'd need to check this again, but from what i found recently.
This should be supported out of the box.
If thats the case we can close this PR

oojacoboo:
@Lappihuan I haven't tried defining the attribute on a constructor parameter/property yet.  I'm not entirely sure if that's considered a property with property promotion or if it's still considered a parameter.  At any rate, we need tests before this can be merged.

PR #402, opened 2021-11-07
Add description and source field name properties to MagicField and SourceField annotations

michael-vostrikov:
MagicField does not have a property for description. SourceField takes description from the source class. It would be good to set property description in the annotation.

When using proxy classes, sometimes it's needed to use different name for a property in the GraphQL type than it is in the source class. If we map one class to another, it would be good to also map properties.

### DIFF
--- a/src/Annotations/MagicField.php
+++ b/src/Annotations/MagicField.php
@@ -37,6 +37,9 @@ class MagicField implements SourceFieldInterface
     /** @var string|null */
     private $description;
 
+    /** @var string|null */
+    private $sourceName;
+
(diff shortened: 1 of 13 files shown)


oojacoboo:
@michael-vostrikov can you please commit a docs update as well.  In particular, the https://graphqlite.thecodingmachine.io/docs/annotations-reference needs updating for `description`.

michael-vostrikov:
@oojacoboo Can you please increment minor version? We would like to pull these changes with composer.

oojacoboo:
@michael-vostrikov done.

PR #403, opened 2021-11-07
Add automatic conversion for DateTimeImmutable and Enum

michael-vostrikov:
We use GraphQLite in Yii 2, where ActiveRecord keeps DATETIME fields as strings. Maybe we can add automatic conversion from string values to DateTimeImmutable?

The same for Enum. Often there is a situation when existing class has enum properties as integers, and other source code contains a logic that depends on it, but it's better to use enums in GraphQL API.

### DIFF
--- a/src/Types/DateTimeType.php
+++ b/src/Types/DateTimeType.php
@@ -25,7 +25,13 @@ class DateTimeType extends ScalarType
      */
     public function serialize($value): string
     {
-        if (! $value instanceof DateTimeImmutable) {
+        if (is_string($value)) {
+            try {
+                $value = new DateTimeImmutable($value);
+            } catch (\Throwable $throwable) {
+                throw new InvariantViolation('DateTime is not an instance of DateTimeImmutable: ' . Utils::printSafe($value), 0, $throwable);
+            }
+        } else if (! $value instanceof DateTimeImmutable) {
             throw new InvariantViolation('DateTime is not an instance of DateTimeImmutable: ' . Utils::printSafe($value));
         }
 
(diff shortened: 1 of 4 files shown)


oojacoboo:
Are you intending on using annotations for informing the parser to convert?  Personally, I'm not a huge fan of this and would rather see it handled another way.

michael-vostrikov:
I would not say this is informing. This is automatic conversion - If output type is set as DateTime and actual value is string, then parser will automatically convert string value to DateTimeImmutable. I don't see real reason to throw exception about types here, GraphQL is not a linter or something, it should not check type correctness in source code on backend. If we manually set output type to be DateTime or some Enum, then we need this type in response, I think the library for GraphQL should provide the way to do this.

Maybe I will add some config setting for this? People who want conversion will enable it, for others it will work as it currently does.

oojacoboo:
@michael-vostrikov I don't think a configuration is necessary here.  Adding this would certainly end up being a bit of a hidden "feature".  There are a couple other ways of solving this issue for you though which I think are better solutions.  

- Use a [custom type mapper](https://graphqlite.thecodingmachine.io/docs/custom-types) (instance of `TheCodingMachine\GraphQLite\Mappers\Root\RootTypeMapperInterface`)
- Use [extended output types](https://graphqlite.thecodingmachine.io/docs/3.0/extend_type)

I'd rather we not adapt a library for use with legacy codebases, using coding practices that are rather dated.  I'm not sure what ActiveRecord ORM you're using, but it should be able to output `DateTimeImmutable` objects instead of strings.  I know I've done this with Propel ORM before.  I realize that means refactoring other code for you, possibly.

michael-vostrikov:
> not adapt a library for use with legacy codebases, using coding practices that are rather dated

Good point, I agree.
Probably custom output types really is better solution here, thanks.

PR #405, opened 2021-11-07
Fix use of MagicField with phpType on a proxy class without magic __get method

### LINKED ISSUE #404: MagicField with phpType on a proxy class without magic __get method
When there is a proxy class for source class with __get method, and proxy class has MagicField with phpType set, the exception is thrown "ReflectionException: Method __get does not exist". This is because of the line `$refMethod = $refClass->getMethod('__get');`, where `$refClass` points to the proxy class, which does not have this method.

michael-vostrikov:
Closes #404

### DIFF
--- a/src/FieldsBuilder.php
+++ b/src/FieldsBuilder.php
@@ -32,6 +32,7 @@ use TheCodingMachine\GraphQLite\Mappers\RecursiveTypeMapperInterface;
 use TheCodingMachine\GraphQLite\Mappers\Root\RootTypeMapperInterface;
 use TheCodingMachine\GraphQLite\Middlewares\FieldHandlerInterface;
 use TheCodingMachine\GraphQLite\Middlewares\FieldMiddlewareInterface;
+use TheCodingMachine\GraphQLite\Middlewares\MissingMagicGetException;
 use TheCodingMachine\GraphQLite\Parameters\InputTypeProperty;
 use TheCodingMachine\GraphQLite\Parameters\ParameterInterface;
 use TheCodingMachine\GraphQLite\Reflection\CachedDocBlockFactory;
@@ -618,8 +619,9 @@ class FieldsBuilder
                 } else {
                     $phpTypeStr = $sourceField->getPhpType();
                     Assert::notNull($phpTypeStr);
-                    $refMethod = $refClass->getMethod('__get');
-                    $type = $this->resolvePhpType($phpTypeStr, $refClass, $refMethod);
+                    $magicGefRefMethod = $this->getMagicGetMethodFromSourceClassOrProxy($refClass);
+
+                    $type = $this->resolvePhpType($phpTypeStr, $refClass, $magicGefRefMethod);
                 }
             }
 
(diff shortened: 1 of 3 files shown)


oojacoboo:
This probably needs to be handled some other way.  Currently, as it's written, if there isn't a `__get` magic method, instead of failing, it's simply assigning `null`.  It's better to fail than to assign something that wasn't intended.

Can you please outline your proxy class example here?

michael-vostrikov:
It is like in test, just a proxy class with MagicField with phpType.
https://github.com/thecodingmachine/graphqlite/pull/405/files#diff-95511f6323fe084d6ab0e77ef3c9c71b20247b1e26c34c0e1d7b0bcf5880d4d0

`$refMethod` is required in `resolvePhpType()` and `mapReturnType()`, so it should be some real method. The right way would be using a `__get()` mehtod from source class, but it is not passed into `getQueryFieldsFromSourceFields()`. Anyway it is used as a stub, real type is taken from `$phpTypeStr`, so I thought a fake method would be good enough.

oojacoboo:
Okay, so after reviewing a bit more, I think I see what you're doing here.  This is an extended type class, right?  It's not exactly a "proxy class"?  Assuming I'm understanding your intent properly, I don't have an issue with getting a proper fix for this use case.  However, the current implementation is quite bad.

Firstly, nested if statements are a big red flag.  But more importantly, the `if (! $sourceField->shouldFetchFromMagicProperty())` evaluation should be returning `false`, letting the magic property logic handle the assignment.

The submitted code hacks something into place that happens to work for your use case.  We need something that works with the existing logic.

michael-vostrikov:
> This is an extended type class, right?

 No, according to [documentation](https://graphqlite.thecodingmachine.io/docs/external-type-declaration#sourcefield-annotation), `TestTypeWithMagicPropertyType` is a proxy class for `TestTypeWithMagicProperty`, just instead of `SourceField` I use `MagicField`.
 
> We need something that works with the existing logic.

Existing logic contains a bug, pull request fixes it. Try to comment changes in `FieldsBuilder` and run the test. You will get PHP exception "ReflectionException: Method __get does not exist", though by documentation this is a valid construction. It works with `MagicField(outputType)` but does not with `MagicField(phpType)`.

oojacoboo:
@michael-vostrikov thanks for the clarity.  Regarding the changes, my comments remain the same.  The existing logic may not work in this case, which I understand.  However, your changes are a hack, and not a good one.  The changes need to be worked into the existing logic in a better way, as I described.

michael-vostrikov:
Found a way how to get source class and magic __get method in it.

oojacoboo:
@michael-vostrikov this is much better!  Can we find a more appropriate method name for `findMagicGetMethodContainer`?  Maybe something like `getRefClassOrProxy`?  Also, it's best to use the conditional check, `if (! $controllerRefClass->hasMethod('__get'))` with an early return.  This keeps the code cleaner.

michael-vostrikov:
I did it a little wrong, `$refClass` should always be a proxy class, because it's where we write class name in phpType, and where we can write `use` for this class. There may not be this `use` in a source class. So the method name now is `getMagicGetMethodFromSourceClassOrProxy`.

oojacoboo:
This looks great @michael-vostrikov.  Thanks for the PR and the follow-up work on it!

Issue #407, opened 2021-11-15
Native Enums Support


enumag:
From the [documentation](https://graphqlite.thecodingmachine.io/docs/type-mapping#enum-types) I can see that this library supports myclabs/php-enum.

The upcoming PHP 8.1 will however have native enums. Are native enums supported as well? If not is there a plan to add the support? I'd like to move away from using myclabs enums if possible.

oojacoboo:
Native enums aren't currently supported.  However, a PR adding support for 8.1 will be happily accepted.  Support for `myclabs/php-enum` will need to remain in place and deprecated until PHP 8.1 has a high enough use percentage.

enumag:
Do you have some estimate of the scope? What would the PR need to implement? Would be helpful if someone wants to try and send such PR. Not saying I'll try. It's possible but can't promise it yet.

oojacoboo:
I'm not familiar with the Enum portion of the lib, but you'd basically just have to translate an Enum to a GraphQL enum type instead of the enum lib.  The lib would need to support both, so you'd just add an additional type translation, basically.

It's probably about 3-4 hours of coding work to get looking good and maybe another hour to get the PR wrapped up.

Lappihuan:
@oojacoboo this can be closed since the PR is merged

PR #409, opened 2021-11-25
Native enum support

dsavina:
This aims to provide support for native enums, introduced with PHP 8.1.

Backed enums using `string` values will base their names on their case values. Non-backed enums, and `int`-backed enums, will instead use their case names.

I'm actually wondering about this last rule: should perhaps the `int`-backed enums be compatible with type `Int` instead of `String`? But then, introspection won't be able to hint allowed values, will it?

Any suggestions?

### DIFF
--- a/src/Mappers/Root/EnumTypeMapper.php
+++ b/src/Mappers/Root/EnumTypeMapper.php
@@ -47,8 +47,12 @@ class EnumTypeMapper implements RootTypeMapperInterface
     /**
      * @param NS[] $namespaces List of namespaces containing enums. Used when searching an enum by name.
      */
-    public function __construct(RootTypeMapperInterface $next, AnnotationReader $annotationReader, CacheInterface $cacheService, array $namespaces)
-    {
+    public function __construct(
+        RootTypeMapperInterface $next,
+        AnnotationReader $annotationReader,
+        CacheInterface $cacheService,
+        array $namespaces
+    ) {
         $this->next = $next;
         $this->annotationReader = $annotationReader;
         $this->cacheService = $cacheService;
@@ -61,8 +65,12 @@ class EnumTypeMapper implements RootTypeMapperInterface
      *
      * @return OutputType&GraphQLType
      */
-    public function toGraphQLOutputType(Type $type, ?OutputType $subType, $reflector, DocBlock $docBlockObj): OutputType
-    {
+    public function toGraphQLOutputType(
+        Type $type,
+        ?OutputType $subType,
+        $reflector,
+        DocBlock $docBlockObj
+    ): OutputType {
         $result = $this->map($type);
         if ($result === null) {
             return $this->next->toGraphQLOutputType($type, $subType, $reflector, $docBlockObj);
@@ -72,12 +80,20 @@ class EnumTypeMapper implements RootTypeMapperInterface
     }
 
     /**
-     * @param (InputType&GraphQLType)|null $subType
+     * Maps into the appropriate InputType
+     *
+     * @param InputType|GraphQLType|null $subType
      * @param ReflectionMethod|ReflectionProperty $reflector
      *
-     * @return InputType&GraphQLType
+     * @return InputType|GraphQLType
      */
-    public function toGraphQLInputType(Type $type, ?InputType $subType, string $argumentName, $reflector, DocBlock $docBlockObj): InputType
+    public function toGraphQLInputType(
+        Type $type,
+        ?InputType $subType,
+        string $argumentName,
+        $reflector,
+        DocBlock $docBlockObj
+    ): InputType
     {
         $result = $this->map($type);
         if ($result === null) {
--- a/src/Mappers/Root/EnumTypeMapper.php
+++ b/src/Mappers/Root/EnumTypeMapper.php
@@ -1,0 +1,198 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TheCodingMachine\GraphQLite\Mappers\Root;
+
+use GraphQL\Type\Definition\InputType;
+use GraphQL\Type\Definition\NamedType;
+use GraphQL\Type\Definition\OutputType;
+use GraphQL\Type\Definition\Type as GraphQLType;
+use phpDocumentor\Reflection\DocBlock;
+use phpDocumentor\Reflection\Type;
+use phpDocumentor\Reflection\Types\Object_;
+use ReflectionClass;
+use ReflectionEnum;
+use ReflectionMethod;
+use ReflectionProperty;
+use Symfony\Contracts\Cache\CacheInterface;
+use TheCodingMachine\GraphQLite\AnnotationReader;
+use TheCodingMachine\GraphQLite\Types\EnumType;
+use TheCodingMachine\GraphQLite\Utils\Namespaces\NS;
+use UnitEnum;
+
+use function assert;
+use function enum_exists;
+
+/**
+ * Maps an enum class to a GraphQL type (only available in PHP>=8.1)
+ */
+class EnumTypeMapper implements RootTypeMapperInterface
+{
+    /** @var array<class-string<UnitEnum>, EnumType> */
+    private $cache = [];
+    /** @var array<string, EnumType> */
+    private $cacheByName = [];
+    /** @var array<string, class-string<UnitEnum>> */
+    private $nameToClassMapping;
+    /** @var RootTypeMapperInterface */
+    private $next;
+    /** @var AnnotationReader */
+    private $annotationReader;
+    /** @var array|NS[] */
+    private $namespaces;
+    /** @var CacheInterface */
+    private $cacheService;
+
+    /**
+     * @param NS[] $namespaces List of namespaces containing enums. Used when searching an enum by name.
+     */
+    public function __construct(RootTypeMapperInterface $next, AnnotationReader $annotationReader, CacheInterface $cacheService, array $namespaces)
+    {
+        $this->next = $next;
+        $this->annotationReader = $annotationReader;
+        $this->cacheService = $cacheService;
+        $this->namespaces = $namespaces;
+    }
+
+    /**
+     * @param (OutputType&GraphQLType)|null $subType
+     * @param ReflectionMethod|ReflectionProperty $reflector
+     *
+     * @return OutputType&GraphQLType
+     */
+    public function toGraphQLOutputType(Type $type, ?OutputType $subType, $reflector, DocBlock $docBlockObj): OutputType
+    {
+        $result = $this->map($type);
+        if ($result === null) {
+            return $this->next->toGraphQLOutputType($type, $subType, $reflector, $docBlockObj);
+        }
+
+        return $result;
+    }
+
+    /**
+     * @param (InputType&GraphQLType)|null $subType
+     * @param ReflectionMethod|ReflectionProperty $reflector
+     *
+     * @return InputType&GraphQLType
+     */
+    public function toGraphQLInputType(Type $type, ?InputType $subType, string $argumentName, $reflector, DocBlock $docBlockObj): InputType
+    {
+        $result = $this->map($type);
+        if ($result === null) {
+            return $this->next->toGraphQLInputType($type, $subType, $argumentName, $reflector, $docBlockObj);
+        }
+
+        return $result;
+    }
+
+    private function map(Type $type): ?EnumType
+    {
+        if (! $type instanceof Object_) {
+            return null;
+        }
+        $fqsen = $type->getFqsen();
+        if ($fqsen === null) {
+            return null;
+        }
+
+        /** @var class-string<object> $enumClass */
+        $enumClass = (string) $fqsen;
+
+        return $this->mapByClassName($enumClass);
+    }
+
+    /**
+     * @param class-string $enumClass
+     */
+    private function mapByClassName(string $enumClass): ?EnumType
+    {
+        if (isset($this->cache[$enumClass])) {
+            return $this->cache[$enumClass];
+        }
+
+        if (! enum_exists($enumClass)) {
+            return null;
+        }
+
+        // phpcs:disable SlevomatCodingStandard.Commenting.InlineDocCommentDeclaration.MissingVariable
+        /** @var class-string<UnitEnum> $enumClass */
+        // phpcs:enable SlevomatCodingStandard.Commenting.InlineDocCommentDeclaration.MissingVariable
+
+        $reflectionEnum = new ReflectionEnum($enumClass);
+
+        $typeAnnotation = $this->annotationReader->getEnumTypeAnnotation($reflectionEnum);
+        $typeName = ($typeAnnotation !== null ? $typeAnnotation->getName() : null) ?? $reflectionEnum->getShortName();
+
+        // Expose values instead of names if specifically configured to and if enum is string-backed
+        $useValues = $typeAnnotation !== null &&
+            $typeAnnotation->useValues() &&
+            $reflectionEnum->isBacked() &&
+            (string) $reflectionEnum->getBackingType() === 'string';
+
+        $type = new EnumType($enumClass, $typeName, $useValues);
+
+        return $this->cacheByName[$typeName] = $this->cache[$enumClass] = $type;
+    }
+
+    private function getTypeName(ReflectionClass $reflectionClass): string
+    {
+        $typeAnnotation = $this->annotationReader->getEnumTypeAnnotation($reflectionClass);
+
+        return ($typeAnnotation !== null ? $typeAnnotation->getName() : null) ?? $reflectionClass->getShortName();
+    }
+
+    /**
+     * Returns a GraphQL type by name.
+     * If this root type mapper can return this type in "toGraphQLOutputType" or "toGraphQLInputType", it should
+     * also map these types by name in the "mapNameToType" method.
+     *
+     * @param string $typeName The name of the GraphQL type
+     */
+    public function mapNameToType(string $typeName): NamedType
+    {
+        // This is a hack to make sure "$schema->assertValid()" returns true.
+        // The mapNameToType will fail if the mapByClassName method was not called before.
+        // This is actually not an issue in real life scenarios where enum types are never queried by type name.
+        if (isset($this->cacheByName[$typeName])) {
+            return $this->cacheByName[$typeName];
+        }
+
+        $nameToClassMapping = $this->getNameToClassMapping();
+        if (isset($this->nameToClassMapping[$typeName])) {
+            $className = $nameToClassMapping[$typeName];
+            $type = $this->mapByClassName($className);
+            assert($type !== null);
+            return $type;
+        }
+
+        return $this->next->mapNameToType($typeName);
+    }
+
+    /**
+     * Go through all classes in the defined namespaces and loads the cache.
+     *
+     * @return array<string, class-string<UnitEnum>>
+     */
+    private function getNameToClassMapping(): array
+    {
+        if ($this->nameToClassMapping === null) {
+            $this->nameToClassMapping = $this->cacheService->get('enum_name_to_class', function () {
+                $nameToClassMapping = [];
+                foreach ($this->namespaces as $ns) {
+                    foreach ($ns->getClassList() as $className => $classRef) {
+                        if (! enum_exists($className)) {
+                            continue;
+                        }
+
+                        $nameToClassMapping[$this->getTypeName($classRef)] = $className;
+                    }
+                }
+                return $nameToClassMapping;
+            });
+        }
+
+        return $this->nameToClassMapping;
+    }
+}
--- a/src/Types/EnumType.php
+++ b/src/Types/EnumType.php
@@ -1,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TheCodingMachine\GraphQLite\Types;
+
+use BackedEnum;
+use GraphQL\Type\Definition\EnumType as BaseEnumType;
+use InvalidArgumentException;
+use UnitEnum;
+
+use function assert;
+use function is_string;
+
+/**
+ * An extension of the EnumType to support native enums.
+ */
+class EnumType extends BaseEnumType
+{
+    /** @var bool */
+    private $useValues;
+
+    /**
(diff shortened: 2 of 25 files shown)


oojacoboo:
Closes #407

oojacoboo:
This looks good @dsavina.  We need to get the tests passing.  Firstly, we'll want to skip the `Enum` tests for version less than 8.1 (https://phpunit.readthedocs.io/en/9.5/incomplete-and-skipped-tests.html#skipping-tests-using-requires).  We should also update the tests to use the latest 8.1 version instead of 8.0.

Regarding `Int`, I think we should be preserving types.  For allowed types, what are you referring to here?  Are you talking about `@Validate` or `@Assertion`?

dsavina:
> we'll want to skip the `Enum` tests for version less than 8.1

I'm aware of the need to filter tests based on PHP version, I'll do it ASAP. In addition, I will also need to isolate enum files in their own namespace to prevent loading invalid PHP code when working with PHP < 8.1.


> Regarding `Int`, I think we should be preserving types. For allowed types, what are you referring to here? Are you talking about `@Validate` or `@Assertion`?

I'm rather talking about the GraphQL standard, which only allows `string` values for an enum. Hence, by construct, it is impossible to fully preserve type for an `int`-backed enum, we will have to chose between declaring a string enum, meaning ignoring backed values, or simply cast to `Int`, meaning not listing available values.

Perhaps an example will clarify, say we have an `int`-backed enum `LogLevel`:
```php
enum LogLevel: int
{
  case Trace = 0;
  case Debug = 1;
  case Info = 2;
  case Warn = 3;
  case Error = 4;
  case Fatal = 5;
}
```
a model `Log`:
```php
class Log
{
  #[Field]
  public LogLevel $level;
  #[Field]
  public DateTime $date;
  #[Field]
  public string $message;
}
```
and a query `logs`:
```php
#[Query]
/** @return Log[] */
public function logs(LogLevel $level = LogLevel::Info): iterable
{
  foreach ($this->logs as $log) {
    if ($log->level->value >= $level->value) {
      yield $log;
    }
  }
}
```

We then have two choices:

- We use case names `Trace`, `Debug`, `Info`, etc. to declare a GraphQL enum type `LogLevel`, available in introspection:
```gql
query {
  _type(name: "LogLevel") {
    enumValues {
      name
    }
  }
}
```
which server would respond with:
```json
{
  "data": {
    "__type": {
      "enumValues": [
        {  
          "name": "Trace"
        },
        {  
          "name": "Debug"
        },
        {  
          "name": "Info"
        },
        ...
      ]
    }
  }
}
```

Requesting `logs` would look like:
```gql
query {
  logs(level: Warn) {
    level
    message
  }
}
```
which server would respond with:
```json
{
  "data": {
    "logs": [
      {
        "level": "Error",
        "message": "Something went wrong"
      },
      {
        "level": "Fatal",
        "message": "Something went really wrong"
      }
    ]
  }
}
```

- We use scalar type `Int`, silently mapping values `0`, `1`, `2`... respectively to `LogLevel::Trace`, `LogLevel::Debug`, `LogLevel::Info`, etc. In this case, any field typed `LogLevel` on server side will appear as a simple `Int` in GraphQL introspection, without hinting available values, but rather throwing exceptions during runtime when receiving out-of-bound values.
Querying logs now looks like this:
```gql
query {
  logs(level: 3) {
    level
    message
  }
}
```
which server would now respond with:
```json
{
  "data": {
    "logs": [
      {
        "level": 4,
        "message": "Something went wrong"
      },
      {
        "level": 5,
        "message": "Something went really wrong"
      }
    ]
  }
}
```
We can also query:
```gql
query {
  logs(level: 6) {
    level
    message
  }
}
```
which would throw an error, as `6` is not a valid `LogLevel` value.

I'm not comfortable with this second option, where the introspection never specified which values are allowed, since parameter `level` is simply signed `Int` for the GraphQL client. I feel like this case should rather be implemented using a custom scalar type.

oojacoboo:
It looks like Apollo Server allows for you to override which one use used through a resolver:
https://www.apollographql.com/docs/apollo-server/schema/schema/#enum-types

From the spec:
>  Enum values are only used in contexts where the precise enumeration type is known. Therefore it’s not necessary to supply an enumeration type name in the literal.

In the case of an `int` value enum, I find that data to generally be important.  In the example you've given, there isn't any way, client side, to know the order of the LogLevels, unless it's been previously agreed upon or following a spec.  However, if the name and value were both provided, there isn't any need to synchronize or converge on some kind of spec.

I can see a lot of cases where, both, the name and value are important.  Is it possible to allow both of these fields while still being compliant with the spec?  I haven't used enums in GraphQL too much.  I do recall the current implementation with `MyCLabs\Enum` being frustrating to work with, for this exact reason.  I believe we ended up swapping the actual values and names in a PHP class, which was not desirable from the backend perspective.

In my opinion, we need to find out if it's possible and/or desirable to output both the name and the value, or we need to
provide an annotation/attribute option to choose which gets output.  And, this should apply to all enums, not just `int` backed ones.

Lappihuan:
There is no such thing as a Enum value in GraphQL, afaik most languages don't expose the Enum value.
Additionally, the API Consumer shouldn't care about the Enum value in PHP, he doesn't know about PHP.

oojacoboo:
@Lappihuan I do not disagree with you and am not suggesting that PHP has any relevance to this conversation.  However, if there is a value that's available and it does not break the spec to include it as a field, I'm not sure I see the issue with providing it.  There are many cases where having a value can be extremely valuable, in conjunction with the name.

If it breaks the spec and won't work, that's that - let's discount the idea, and move forward with an annotation option.

oojacoboo:
@dsavina would be great to get this wrapped up.  We should go ahead with what you have now. 

In order for us to support the option to output either the constant name or the assigned value, we'd really need to add a new attribute for Enum types, or possibly add a new property to the `Type` attribute (don't like this).  This can be added as a new feature improvement.

As for outputting the name and the value, I don't even know if you could query the "value" of an Enum through most clients.  Technically you could make it available in the `__schema`.  However, that doesn't mean clients would allow access, as many are probably just assigning a single string value.

aszenz:
If I understand the issue correctly, why can't we only allow string backed enums and refuse to accept other php enums ?
That seems to me the simplest option

Lappihuan:
i'm still wondering why this is even a question.
the library already implements enums with the 3rd party library myclabs/php-enum.
a 1:1 implementation of that with native enums is the only thing that makes sense.

i am still confused why the value of a enum would be of any intrest for an api consumer.

the current implementation would work like this:

PHP:
`
class StatusEnum extends Enum
{
    private const ON = 'on';
    private const OFF = 'off';
    private const PENDING = 'pending';
}
`
GQL:
`
enum StatusEnum {
    ON
    OFF
    PENDING
}
`
No mather what the backed value is:
PHP:
`
class StatusEnum extends Enum
{
    private const ON = 1;
    private const OFF = 2;
    private const PENDING = 3;
}
`
GQL:
`
enum StatusEnum {
    ON
    OFF
    PENDING
}
`

oojacoboo:
We should just follow the webonyx implementation.

https://webonyx.github.io/graphql-php/type-definitions/enums/

The current `MyCLabs` implementation uses the key for both the name and value of the `EnumType`, internally.  This results in the resolver always returning the key/name.  I know I've ran into this issue before during resolution.  Internally, you want the value assigned, if it differs from the key/name.

I'm all for getting this right now though, now, and not trying to create a 1:1 implementation of the `MyCLabs` implementation.

Current `MyCLabs` implementation:

```php
class MyCLabsEnumType extends EnumType
{
    public function __construct(string $enumClassName, string $typeName)
    {
        $consts         = $enumClassName::toArray();
        $constInstances = [];
        foreach ($consts as $key => $value) {
            $constInstances[$key] = ['value' => $enumClassName::$key()];
        }

        parent::__construct([
            'name' => $typeName,
            'values' => $constInstances,
        ]);
    }
...
```

As you can see, the key is being assigned to both, the key and value.  However, the webonyx implementation supports keys and values, according to the native `BackedEnum` and resolves them accordingly.

Lappihuan:
Exposing the value over GQL is just not in the GQL Spec...
As for mapping return types internally, that has nothing to do with GQL but just the TypeMapper.
The TypeMapper should be designed to allow all usecases of native enums.

oojacoboo:
@Lappihuan can you be more concrete here in terms of what you're suggesting different?  What is wrong with implementing Enums according to the webonyx type mapping implementation?

dsavina:
Sorry I couldn't get back on this subject earlier.

&nbsp;

> @oojacoboo commented 5 days ago
> As you can see, the key is being assigned to both, the key and value.

This is not exact: the enum key is indeed used as the key in the Webonyx representation of the enum case. However, the value obtained with expression `$enumClassName::$key()` is an instance of `TEnum of MyCLabs\Enum\Enum`. If I understand correctly, Webonyx exposes the key to the GraphQL Api, and map it internally to the corresponding value (i.e. an object, instance of the corresponding enum class, whether `MyCLabs` or native).

Therefore, the question remains regarding the exposed names. I think the best approach would be to leave the choice to developers, by supporting an option `mapValues` in the `Type` annotation, only applicable for string-backed enums. Default behavior would map the enum names.

oojacoboo:
@dsavina I'm all for offering the option here.  That was my initial recommendation.  For the sake of simplicity, I think we could go with a direct mapping according to the `webonyx` implementation.

As for the option, we have the `EnumType` attribute used to change the name of an Enum type.  Right now, Enums are converted to Types by extending the `MyCLabs` enum class, via the type-mapper.  I'm assuming this is the reason for the additional attribute, as opposed to using the default `Type` attribute.

I think we should unify these attributes by deprecating the `EnumType` attribute, as the `Type.name` will serve for this purpose.  Then, we can add an argument to the `Type` attribute, clarifying that it's intended for Enums - maybe `bool $useEnumValues`.  `mapEnumValues` feels a bit ambiguous, but I don't hate it.

oojacoboo:
Hey @dsavina I'd like to target a 5.1 release with this and #435.  Any chance you'll have some time to get this wrapped up soon?

dsavina:
Hi @oojacoboo,

I'm sorry, I'm very busy these days, I won't be able to come back on this for the next two weeks. I should find the time to do it somewhere around February 10, would that work?

oojacoboo:
@dsavina that should be fine.  We'll be anxiously awaiting :)

withinboredom:
> We'll be anxiously awaiting :)

image

dsavina:
Hi there,

Just a quick update: I found some time to change the implementation in order to use `Type` annotation for enums, with optional argument `useEnumValues` to expose values instead of names.

I couldn't update the tests in order to bypass all this whenever running under 8.1, I will do it first thing on Monday.

dsavina:
Hi,

I did all I could, I'm not 100% satisfied but I guess it can be improved afterwards:
- I had to rollback on using `Type` annotation instead of `EnumType`, as it somehow led to the schema trying to declare an object type conflicting with the already existing enum type.
- I had to disable PHPStan on the files referencing PHP 8.1-specificities (`UnitEnum`, `BackedEnum`, `enum_exists`, `ReflectionEnum`), since the analysis failed with versions prior to 8.1.
- Tests are OK whether running with 8.1 or older; I did some gross conditional service declaration to that end, but no shame there :)

src/Types/EnumType.php
```diff
@@ -0,0 +1,63 @@
...
+     * @return string
+     */
+    public function serialize($value)
+    {
+        if (! $this->useValues) {
+            return $value->name;
+        }
+
+        assert($value instanceof BackedEnum);
+        assert(is_string($value->value));
```
oojacoboo, line 56: While assert is pretty cool, being a configurable option, I think it's best to throw Exceptions here.

src/Mappers/Root/EnumTypeMapper.php
```diff
@@ -0,0 +1,199 @@
...
+            return $this->cache[$enumClass];
+        }
+
+        if (! enum_exists($enumClass)) {
+            return null;
+        }
+
+        // phpcs:disable SlevomatCodingStandard.Commenting.InlineDocCommentDeclaration.MissingVariable
+        /** @var class-string<UnitEnum> $enumClass */
+        // phpcs:enable SlevomatCodingStandard.Commenting.InlineDocCommentDeclaration.MissingVariable
```
oojacoboo, line 119: Are we getting much value out of this?  I'd just remove it.

src/Mappers/Root/EnumTypeMapper.php
```diff
@@ -0,0 +1,199 @@
...
+            $type = $this->mapByClassName($className);
+            assert($type !== null);
+            return $type;
+        }
+
+        return $this->next->mapNameToType($typeName);
+    }
+
+    /** @var array<string, class-string<UnitEnum>> */
+    private $nameToClassMapping;
```
oojacoboo, line 172: Let's put all properties at the top of the class.

oojacoboo:
@dsavina thanks again for the work on this.  I added a few comments on the code, but overall I think this is good.  

I agree that it'd be ideal to use a `Type` annotation instead of a custom `EnumType`.  Deprecating `EnumType` now would allow us to tie that directly to the use of the `MyCLabs` enum.  If we support `EnumType` now for `BackedUnum` instances, we won't be able to remove it at the same time as `MyCLabs` support - in the future.

Did you follow the implementation for the `MyCLabs` enum?  Does it have the same object type declaration issue?

src/Mappers/Root/EnumTypeMapper.php
dsavina: PHPStan (before I disabled it) required a `class-string<UnitEnum>`, as signed in `EnumType` constructor, and wasn't able to determine `$enumClass` complies at this point (after the `enum_exist` check).

On the other side, CodeSniffer doesn't like precising the type subsequently to a variable declaration (which I think is a shame, but I guess I understand the rule), therefore the `phpcs:disable` directive.

I could simply soften the typecheck in `EnumType::__constructor`, from `class-string<UnitEnum>` to `class-string`, but I must say I'm not so fond of this option.

src/Types/EnumType.php
dsavina: `EnumTypeMapper` actually validated these two assertions:

```php
        // Expose values instead of names if specifically configured to and if enum is string-backed
        $useValues = $typeAnnotation !== null &&
            $typeAnnotation->useValues() &&
            $reflectionEnum->isBacked() &&
            (string) $reflectionEnum->getBackingType() === 'string';
```

Unless we don't trust the core to input values according to the mapper? (in which case I should sign `mixed` and also make sure `$value instanceof UnitEnum` in the first place)

dsavina:
I'm not sure I understand the last question, but my guess is that using `Type` instead of `EnumType` on `MyCLabs` enums would raise the same errors.

I suspect the `RecursiveTypeMapper` to priorize the mappers differently and pull the rug under the `EnumTypeMapper`, but this is pure speculation, I have no proof for a strong case here.

src/Types/EnumType.php
oojacoboo: That's a good point.  I wasn't aware checks were being done there.  I'd still argue that the requirements of one class shouldn't be reliant on another.  ideally we'd rely on proper type checking for this, but we're still supporting 7.2 at the moment.  We really need to bump the min support version to 7.4 IMO.

We can leave this - proper type checking should resolve in the future.

src/Mappers/Root/EnumTypeMapper.php
oojacoboo: IMO, we should be doing better validation in the constructor for this string, instead of relying on static type analysis.  What kind of feedback is a developer going to here here if an incorrect class-string is provided?  It's nice to provide these additional static checks for IDE completion and lib development testing, but provides little runtime, or even development, assurances.

Is this error not referring to the fact that it's a `param` and not a `var` in terms of phpcs parsing?  Why aren't we applying this in the docblock?

oojacoboo:
So, regarding my last comment, it's mostly around future deprecation and API planning.  If we support `EnumType` for `UnitEnum` now, it'll be much harder to deprecate support for `EnumType`, ever - since people will be relying on it for enums built into PHP.  Whereas, deprecating `EnumType` now, targeting removal in the near future, isn't such a hard thing to do because you could migrate to using `Type` and PHP enums, instead of `MyCLabs`.

dsavina:
As I said earlier:
> I had to rollback on using Type annotation instead of EnumType, as it somehow led to the schema trying to declare an object type conflicting with the already existing enum type.


To give you more context, my first implementation using `Type` annotations resulted in the schema analyzer confusing enum fields as object types when running `Schema::assertValid()`:
```diff
- GraphQL\Error\InvariantViolation : Schema must contain unique named types but contains multiple types named "Color" (see http://webonyx.github.io/graphql-php/type-system/#type-registry).
```

Debugging showed that, indeed, in `GraphQL\Utils\TypeInfo::extractTypes()`, assertion `$typeMap[$type->name] === $type` fails, since `$type` is a `TypeAnnotatedObjectType`. Sadly, my understanding ends here, I'm clearly not familiar enough with GraphQLite, nor do I have the time to investigate this thoroughly :confused: Perhaps you or @moufmouf will have an idea?

oojacoboo:
@dsavina I don't think @moufmouf monitors this lib anymore.  I haven't seen any responses from him in quite some time.

Regarding that error, that's a fairly common one.  My guess is that a type is being created according to the implementation you had, but also it's assumed, as you state, that it's a standard type and creating one for that as well.  It shouldn't be too difficult to add a condition, check if it's an enum, and push the mapping accordingly.

Do you still have the implementation for this?  If you can commit that, I'd be happy to have a look - see if I can get it working.

dsavina:
@oojacoboo, I pushed branch [test/native-enum-support](https://github.com/dsavina/graphqlite/tree/test/native-enum-support) on my own repository, if you want to take a look!

dsavina:
Hi @oojacoboo, did you manage to make this work?

oojacoboo:
@dsavina I haven't had a chance to look into it yet.  I plan to do so very soon.

dsavina:
@oojacoboo, any news? It can be hard to find time to work on open source projects, I totally get it as I go through the exact same struggle.

Nevertheless, this Merge Request has been open for four months now, and I'm worried that it is going to be simply abandoned.

It seems that deprecating `EnumType` has trickier impacts than it looked in the first place, perhaps it would need proper reflexion and discussion in a separate task.

Pragmatically, I think it's smarter to allow the use of something we might deprecate in the future, than to simply block the evolution of the project due to expectations higher than we can handle. What do you think? Could we merge this as it is, and defer the deprecation of `EnumType` in a dedicated Merge Request afterwards (that is, if it turns out to be possible)?

Lappihuan:
> As I said earlier:
> 
> > I had to rollback on using Type annotation instead of EnumType, as it somehow led to the schema trying to declare an object type conflicting with the already existing enum type.
> 
> To give you more context, my first implementation using `Type` annotations resulted in the schema analyzer confusing enum fields as object types when running `Schema::assertValid()`:
> 
> ```diff
> - GraphQL\Error\InvariantViolation : Schema must contain unique named types but contains multiple types named "Color" (see http://webonyx.github.io/graphql-php/type-system/#type-registry).
> ```
> 
> Debugging showed that, indeed, in `GraphQL\Utils\TypeInfo::extractTypes()`, assertion `$typeMap[$type->name] === $type` fails, since `$type` is a `TypeAnnotatedObjectType`, vs `$typeMap[$type->name]` being the expected `EnumType`. Sadly, my understanding ends here, I'm clearly not familiar enough with GraphQLite, nor do I have the time to investigate this thoroughly 😕 Perhaps you or @moufmouf will have an idea?

This sounds like some other typemapper already felt responsible to resolve the enum.
I haven't looked at the native enum implementation but maybe PHPs reflection allows to treat them as normal classes too?

From this bit in the docu i think other typemappers would need to return null if they are called on a native enum:
https://graphqlite.thecodingmachine.io/docs/custom-types#registering-a-custom-scalar-type-advanced

Alternativley the Native Enum typemapper could be called first in the chain.

As for merging this PR.
why is tying the EnumType to MyClabs even that important for deprecation, its a internal class.

oojacoboo:
@dsavina can you allow commits on this branch ([test/native-enum-support](https://github.com/dsavina/graphqlite/tree/test/native-enum-support))? I've been able to get tests passing where they were failing. There is one other test failing due to a bad graphql query. But, other than that, all other tests are passing. You can have a look at what I put in place. I'm not convinced that the logic is in the best spot, or the approach is ideal. But it is working.

dsavina:
@oojacoboo, great news, thanks for the update!

You can push directly here on [feature/native-enum-support](https://github.com/dsavina/graphqlite/tree/feature/native-enum-support) if it's working, maintainers edits are allowed.

If you prefer that I take a look before pushing here, I guess the simplest is that you push on your repository :)

oojacoboo:
@dsavina I'm not sure what changes you've made on this branch to support the `EnumType` annotation for native enums.  I'd think we'd be left with `EnumType` fragments if I pushed on here, no?

oojacoboo:
@dsavina I've merged and pushed everything onto this PR.  As you can see, there is a failing test for an enum.  It looks like it's not using the values of the enum properly.  If you could please resolve this issue and review everything else, we can go ahead and get this merged in ASAP - target a new release.

oojacoboo:
@dsavina there was a bug with the `useEnumValues` property not registering.  That bug has been resolved and tests are now passing.  This PR is ready to merge.  Please review and let me know on the previous questions

>Questions:
>
>How does the class property of the Type annotation apply with enums? We probably should update annotation >documentation on this.
>~~Do we have tests covering the @Type annotation with different properties being applied on enums?~~

oojacoboo:
I've updated the documentation a bit and it's now merged into master.  Thanks for kicking things off with this @dsavina and everyone else for your contributions.

dsavina:
Hi @oojacoboo, sorry for not being available during the past week. This is great news, thank you for the last corrections!

oojacoboo:
For the record, I've been able to test this on a rather large API implementation and everything looks good.

porozhnyy:
Hey! Thanks for the great changes! Are there any plans to make a new release with these changes? @oojacoboo @dsavina

oojacoboo:
Please see here:
https://github.com/thecodingmachine/graphqlite/discussions/469#discussioncomment-2631955

PR #446, opened 2022-03-01
upgrade phpstan 1.x

janatjak:


src/SchemaFactory.php
```diff
@@ -391,7 +391,7 @@ public function createSchema(): Schema
         $inputTypeUtils     = new InputTypeUtils($annotationReader, $namingStrategy);
         $inputTypeGenerator = new InputTypeGenerator($inputTypeUtils, $fieldsBuilder);
 
-        if (empty($this->typeNamespaces) && empty($this->typeMappers)) {
+        if (empty($this->typeNamespaces) && empty($this->typeMappers) && empty($this->typeMapperFactories)) {
```
oojacoboo, line 394: Why are we adding an additional condition here for throwing an exception?  This is relaxing the state where an exception is thrown, requiring that the `typeMapperFactories` also evaluate to false.

src/Types/MutableInputObjectType.php
```diff
@@ -90,6 +91,7 @@ public function getFields(): array
                     if ($fieldDefinition instanceof Type) {
                         $fieldDefinition = ['type' => $fieldDefinition];
                     }
+                    Assert::string($name);
```
oojacoboo, line 94: Why are we using this lib?  What's wrong with `assert(is_string($name))`?

src/Types/MutableTrait.php
```diff
@@ -85,6 +86,7 @@ public function hasField($name): bool
     public function getFields(): array
     {
         $this->initializeFields();
+        Assert::isArray($this->fields);
```
oojacoboo, line 89: Another use of `Assert` lib - why?

oojacoboo:
Thanks for the PR - wasn't expecting to see this much logic refactoring, but it looks good for the most part, certainly clears up some of the existing logic in places.  I added some comments if you could please address those.

phpstan.neon
```diff
@@ -1,10 +1,10 @@
-#includes:
+includes:
 #	- vendor/phpstan/phpstan/conf/bleedingEdge.neon TODO: Try to add this
 parameters:
     tmpDir: .phpstan-cache
     paths:
         - src
-    level: max
+    level: 8
```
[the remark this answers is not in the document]
janatjak: Level 9 contains 116 new errors -> I decided not to use it now

src/Containers/BasicAutoWiringContainer.php
```diff
@@ -20,7 +19,7 @@ class BasicAutoWiringContainer implements ContainerInterface
 {
     /** @var ContainerInterface */
     private $container;
-    /** @var ObjectType[] */
+    /** @var mixed[] */
```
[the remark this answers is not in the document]
janatjak: If you look closely on `get` method 

https://github.com/thecodingmachine/graphqlite/blob/624e70932d33f3bbfcbb432e970938632b08f69e/src/Containers/BasicAutoWiringContainer.php#L60

values can contain anything

src/SchemaFactory.php
janatjak: I wanna use SchemaFactory only with `addTypeMapperFactory`, without calling addTypeMapper.

src/Types/MutableInputObjectType.php
janatjak: fixed

src/Types/MutableTrait.php
janatjak: fixed

src/SchemaFactory.php
oojacoboo: I don't think this is the right way to go about this.  Is a factory a requirement for the `SchemaFactory` to work properly?  Why can't i just pass in `typeNamespaces` and `typeMappers` only?  Keep in mind that your commits apply to everyone using this lib and not everyone is going to use it the same.

I'm still not quite sure what you're trying to accomplish here by throwing an exception if the factory isn't present.

src/Containers/BasicAutoWiringContainer.php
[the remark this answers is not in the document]
oojacoboo: The logic of the `get` method looks to assume it's an object.  I don't use this AutoWiring and never would.  So, I'm not entirely sure all the inner-workings.  But `mixed` seems inappropriate here, unless I'm missing something.  `ObjectType` also seems incorrect, if this thing is a generic auto wiring attribute/service.  Assuming so, maybe `object` is the correct type?

src/SchemaFactory.php
oojacoboo: If you intend on only relying on the factories, maybe the better logic is to check to see if there are factories present, process those to get the type mappers, assigning them to the `typeMappers` property, and then leaving the exception as it is, checking to see if `typeMappers` and `typeNamespaces` are both not empty.

janatjak: IMHO this is valid example, how i want to use this lib. But its not possible now.

```php
    $mapperFactory = new StaticClassListTypeMapperFactory($classList);
    $queryProviderFactory = new AggregateControllerQueryProviderFactory($controllers, $container);
    $factory = new SchemaFactory($cache, $container);
    $factory->addQueryProviderFactory($queryProviderFactory);
    $factory->addTypeMapperFactory($mapperFactory);
    $schema = $factory->createSchema();
```

oojacoboo: @janatjak I think that's a valid expectation.  Please see my previous comment.  The logic here needs to be written to be flexible and cover all scenarios, not just this one.  In your code, if I do not include a factory, I'll get an exception and it shouldn't be a requirement to include a factory, right?

src/Containers/BasicAutoWiringContainer.php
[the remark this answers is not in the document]
janatjak: agree, you are right

`object` is the best choice

src/SchemaFactory.php
janatjak: I'm sorry, I missed your prev comment. Pls check my last commit. Thanks

oojacoboo: Why not process the factories first and add the type mappers from the factory to the class property `typeMappers` instead of creating an additional variable?  The logic looks good though, just could be a little cleaner I think.

oojacoboo: Doing what I'm suggesting would mean you'd only need to add the `$compositeTypeMapper->addTypeMapper($typeMapper);` call within the loop on `typeMappers`.

janatjak: good point, thank you for your hint

oojacoboo:
@janatjak One last question before we merge this.  Do we have test coverage on the `SchemaFactory` testing the scenario where only factories are provided and only mappers provided?

janatjak:
> @janatjak One last question before we merge this. Do we have test coverage on the `SchemaFactory` testing the scenario where only factories are provided and only mappers provided?

Done

oojacoboo:
Thanks again for working through this.

PR #455, opened 2022-04-04
Add config example with PHP 8 attributes to input types docs

andrew-demb:
Unfortunately, I don't know how to check the built result.

Copied tabs markup from example from the same page

### DIFF
--- a/website/docs/input-types.mdx
+++ b/website/docs/input-types.mdx
@@ -412,11 +412,29 @@ public function createLocation(float $latitude, float $longitude): Location
 
 In case you want to override the input type name, you can use the "name" attribute of the @Factory annotation:
 
+<Tabs
+  defaultValue="php8"
+  values={[
+    {label: 'PHP 8', value: 'php8'},
+    {label: 'PHP 7', value: 'php7'},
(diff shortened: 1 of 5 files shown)


oojacoboo:
@andrew-demb thank you for these efforts.  We're overdue for a release.  However, I believe you can check the "Next" version on the docs to see attribute examples that have already been added/updated.

oojacoboo:
We could merge this though, if you'd like to confirm.  On your local machine, you can run `npm run serve` I believe it is and it'll run an npm server with the built docs.

andrew-demb:
Fixed introduced inconsistency between annotation and attribute example.

I can confirm result is displayed correctly.
Should I add the same changes to all versioned docs too (as of now it is applied only for the 'next' release)?

<details>
  <summary>Screenshot</summary>

image
</details>

oojacoboo:
@andrew-demb Thanks.  You could also update for >= 4.1 archives as well.  I believe, 4.1 was the first version to support attributes.

andrew-demb:
Checked and done.
Thanks for your help.

PR #458, opened 2022-04-09
#[Input] Type Improvements

oojacoboo:
This PR started out by addressing an issue with nested InputTypes not being resolved and needing support within the `RecursiveTypeMapper`.  As this was resolved, other issues came to light and were resolved as well:

- Nested InputTypes are now supported, regardless of their loading order in type mappers.  So, if InputType `A` references InputType `B` , `B` will be resolved as well, even if it hasn't prior.  This is what the `RecursiveTypeMapper` does, only it was working with output Types.
- When the `Input` annotation was released, there was a known issue where a `Type` annotation also had to be used in some cases.  This issue should be resolved now.
- Input types are receiving some additional validation as part of their addition to the `RecursiveTypeMapper`

The current `Input` implementation is a bit shaky.  This PR provides more stability and better integration into type mapper processing.

### DIFF
--- a/src/AnnotationReader.php
+++ b/src/AnnotationReader.php
@@ -26,6 +26,7 @@ use TheCodingMachine\GraphQLite\Annotations\ParameterAnnotationInterface;
 use TheCodingMachine\GraphQLite\Annotations\ParameterAnnotations;
 use TheCodingMachine\GraphQLite\Annotations\SourceFieldInterface;
 use TheCodingMachine\GraphQLite\Annotations\Type;
+use TheCodingMachine\GraphQLite\Annotations\TypeInterface;
 use Webmozart\Assert\Assert;
 
 use function array_diff_key;
@@ -262,7 +263,7 @@ class AnnotationReader
      *
      * @template T of object
      */
-    public function getTypeAnnotation(ReflectionClass $refClass): ?Type
+    public function getTypeAnnotation(ReflectionClass $refClass): ?TypeInterface
     {
         try {
             $type = $this->getClassAnnotation($refClass, Type::class)
--- a/src/AnnotationReader.php
+++ b/src/AnnotationReader.php
@@ -93,207 +93,6 @@ class AnnotationReader
     }
 
     /**
-     * @param ReflectionClass<T> $refClass
-     *
-     * @template T of object
-     */
-    public function getTypeAnnotation(ReflectionClass $refClass): ?Type
-    {
-        try {
-            $type = $this->getClassAnnotation($refClass, Type::class);
-            if ($type !== null && $type->isSelfType()) {
-                $type->setClass($refClass->getName());
-            }
-        } catch (ClassNotFoundException $e) {
-            throw ClassNotFoundException::wrapException($e, $refClass->getName());
-        }
-
-        return $type;
-    }
-
-    /**
-     * @param ReflectionClass<T> $refClass
-     *
-     * @return Input[]
-     *
-     * @throws AnnotationException
-     *
-     * @template T of object
-     */
-    public function getInputAnnotations(ReflectionClass $refClass): array
-    {
-        try {
-            /** @var Input[] $inputs */
-            $inputs = $this->getClassAnnotations($refClass, Input::class, false);
-            foreach ($inputs as $input) {
-                $input->setClass($refClass->getName());
-            }
-        } catch (ClassNotFoundException $e) {
-            throw ClassNotFoundException::wrapException($e, $refClass->getName());
-        }
-
-        return $inputs;
-    }
-
-    /**
-     * @param ReflectionClass<T> $refClass
-     *
-     * @template T of object
-     */
-    public function getExtendTypeAnnotation(ReflectionClass $refClass): ?ExtendType
-    {
-        try {
-            $extendType = $this->getClassAnnotation($refClass, ExtendType::class);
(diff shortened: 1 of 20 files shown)


Lappihuan:
We are currently using deeply nested Input Types and have no issue with it.
Would you mind showing a repo of the issues?

Also setting default to true sounds like a bad change.

oojacoboo:
@Lappihuan The nested InputTypes you have, using the `@Input` annotation, are they perhaps sorted such that the Input type mapper processes them first in a directory?  Also, another issue was related to standard output Types being referenced in an InputType.  Can you give this PR a test on your API?

The default setting follows the `@Type` implementation.  Without being defined as default, the type is skipped over in the `GlobTypeMapperCache::78`.  Defaulting to true is a sane default.  Why wouldn't `#[Input]` be considered the default of this type?  This is only a BC break if you need the following:

```php
#[Input(name: 'SomethingInput']
#[Input(name: 'UpdateSomethingInput', default: false]
```

I will agree that it's nicer to define the truthy default from an API perspective, but the input annotations are not processed in such a way that we can easily determine if there are multiple on a single class/type.

Lappihuan:
> @Lappihuan The nested InputTypes you have, using the `@Input` annotation, are they perhaps sorted such that the Input type mapper processes them first in a directory? Also, another issue was related to an enum Type (using `@Type`) being referenced in an InputType as a field return value.

We don't do anything special with them, they work as long as they are in the configured TypeNamespace.
This is a typical Type we use for both input/output with deeply nested inputs and enum values:
```php
#[GQL\Type(default: true)]
#[GQL\Input(name: 'CreateMailInput', default: true)]
#[GQL\Input(name: 'UpdateMailInput')]
class Mail
{

    #[GQL\Field(for: 'CreateMailInput', inputType: 'CreateMailArchiveInput!')]
    #[GQL\Field(for: 'UpdateMailInput', inputType: 'UpdateMailArchiveInput')]
    public ?MailArchive $mailArchive = null;

//...

    #[GQL\Field]
    public function getMailArchive(#[GQL\Autowire(identifier: "app.serializer")] SerializerInterface $serializer): ?MailArchive
    {
        return $serializer->denormalize($this->soapData, MailArchive::class, "array", [
            AbstractObjectNormalizer::DISABLE_TYPE_ENFORCEMENT => true,
            "groups" => 'mailarchive'
        ]);
    }
//...
```
and the subtype MailArchive handling an Enum:
```php
#[GQL\Type(default: true)]
#[GQL\Input(name: 'CreateMailArchiveInput', default: true)]
#[GQL\Input(name: 'UpdateMailArchiveInput')]
class MailArchive
{
    /**
     * @var ?MailArchiveExclusionSensitivtyEnum[]|null
     */
    #[GQL\Field]
    #[Groups('mailarchive')]
    #[SerializedName("mailarchiveexclusionsensitivity")]
    public ?array $mailArchiveExclusionSensitivity = null;

//...

    /**
     * @return ?MailArchiveExclusionSensitivtyEnum[]|null
     */
    public function getMailArchiveExclusionSensitivity(): ?array
    {
        return $this->mailArchiveExclusionSensitivity;
    }
```

> Can you give this PR a test on your API?

how would i do that? set the branch as version in the composer.json?

> I've taken another look at the `default` implementation and have reverted the changes. It's not something I was fond of anyway. It's worth noting that the usage may not be inherently clear. I'll add some notes to the documentation as well. When defining multiple `@Input` annotations, if a `name` attribute is provided, if will automatically default to `false`. This is a bit strange, but it provides for a cleaner API. If you provide a name, you must also provide a `default` attribute if you want the InputType to be the "default".
> 
> ```
> #[Input] // Is default
> #[Input(default: false)] // Not default
> #[Input(name: 'SomethingInput', default: true] // Is default
> #[Input(name: 'UpdateSomethingInput'] // Not default
> ```
As you can see in the example above we use two defaults, one for Query (Type) and one for Mutation (Input). The current implementation makes sense that you define the default as soon as its not clear which one is the default implicitly

oojacoboo:
@Lappihuan thanks for sharing those examples.  Since you're using the `@Type` annotation as well, the recursive type mapper will already handle the return type mapping for all fields.  Without the `@Type` annotation is where you run into issues.

To give the PR a try, yea - add it to your `composer.json`.  To do so, you'll first want to add the repository to your `repositories` array at the top level of the config:

```json
{
    ...
    "repositories": [
        {
            "type": "github",
            "url": "git@github.com:oojacoboo/graphqlite.git"
        }
    ]
}
```

Then, for your require, simply reference the branch prefixed by `dev-`:

```json
{
    ...
    "require": {
        "thecodingmachine/graphqlite": "dev-recursive-input-types"
    }
}
```

Lappihuan:
@oojacoboo we did test your branch on our API and it seems to be working fine.

related to this we were looking at making our input types better and one issue that came up was setters not behaving like expected.
We have made a fork of your branch here and added functionality which brings setters closer to how getters work on Types.
There is some things still open to fix, which we will in the next days.

[Here](https://github.com/Lappihuan/graphqlite/tree/bring_input_setters_on_par_with_type_getters) have a look if you like.

oojacoboo:
Thanks @Lappihuan for the review.

Yea, I think the input types were originally written with the expectation that the properties would be hydrated directly.  You can also use the constructor which is useful for validation and other pre-processing needs.

I'm going to merge this into master now and look forward to reviewing your PR.

Issue #449, opened 2022-03-20
Circular dependencies in type mapping cause seg fault using `Input` attribute


oojacoboo:
When the `Input` type mapper is building out the schema, if there are circular dependencies between types, you'll end up with a seg fault.

We need to review the type mapper logic and see if it's possible to predict the existence of a circular dependency and throw an exception or find a way to support them, if it's possible to have the logic look back through already processed types and see if the one being processed has been processed prior.  If so, ignore it or associate it with the type being processed as a relation (if necessary).

The current issue exists where I have the following:

```php
#[Input]
class Foo
{
    /**
     * @var Bar[]
     */
    #[Field]
    protected array $bars;
```

```php
#[Input]
class Bar
{
    #[Field]
    protected Baz $baz;
```

```php
#[Input]
class Baz
{
    #[Field]
    protected Foo $foo;
```

oojacoboo:
@devmaslov any thoughts on this?  You wrote this `Input` type mapper implementation.

Lappihuan:
With #466 Input Types use callables too which could fix this issue.
[explanation](https://webonyx.github.io/graphql-php/type-definitions/object-types/#recurring-and-circular-types) from graphql-php

Lappihuan:
I actually think its that we try to get all fields before we pack the configs in a callable, since the old implementation also used a callable:
https://github.com/thecodingmachine/graphqlite/pull/466/files#diff-801f55b45a3306c52e2892331cc1e8cb5c41f5800b9d8c089d37d5e07b7c5a99L48-R49

I'm not sure how to solve getting the fields only inside the callable but also still set them on the class property to be used later in resolve...

Lappihuan:
This should be fixed with https://github.com/thecodingmachine/graphqlite/pull/466/commits/ef0c53119b91ad3545e90d3b4b121eafebddd318
@oojacoboo can you try to confirm this?

oojacoboo:
@Lappihuan nice.  I don't actually have any ability to test this.  I looked over the commit and it looks good.  Is this part of #466?

Lappihuan:
yes, its in #466 
i built a test that goes as far as building the schema, since your description said thats where the segfault happend.
https://github.com/thecodingmachine/graphqlite/pull/466/files#diff-eaf66d30e18cc3db1f122fd089ce3ce52deb80fdb44218807bd3964d67e21394R2385-R2397

oojacoboo:
[...]

Issue #371, opened 2021-08-26
The inputType "String!" of a field in an annotated input type is being ignored when update is set to true.


hjerichen:
When an input type ist specified with `update: true`, it makes all fields optional (as stated in the documentation).
It even does it when I specifically say it should be non optional through the inputType in a Field Annotation.

```php
#[Input(update: true)]
class UserInput
{
    #[Field(inputType: 'String!')]
    public string $important;
}
```
The output will be `String` and not `String!`.

Is this intended behavior?
I would expect that my specified input type is used.

Lappihuan:
As explained by @devmaslov [here](https://github.com/thecodingmachine/graphqlite/pull/269#issue-634866663), this is the intended behavior.
I will check if its easy to implement.

oojacoboo:
@Lappihuan The example given here should default to `String!` anyway, since it's not nullable typed in PHP, regardless of the annotation.

Lappihuan:
@oojacoboo no, update makes everything nullable regardless of what php says.
this is actually intended and a really cool feature as described [here](https://graphqlite.thecodingmachine.io/docs/input-types#multiple-input-types-per-one-class)
```
Note that update: true argument for UpdateUserInput. It should be used when input type is used for a partial update, It makes all fields optional and removes all default values from thus prevents setting default values via setters or directly to public properties. In example above if you use the class as UpdateUserInput and set only username the other ones will be ignored. In PHP 7 they will be set to null, while in PHP 8 they will be in not initialized state - this can be used as a trick to check if user actually passed a value for a certain field.
```

oojacoboo:
@Lappihuan yep, overlooked that it was an `update` here.  I guess this is the right approach here.  I do wonder if we'd benefit from another attribute on the `Field` annotation to make it "required".  Or maybe that's at least worth looking into/considering.

Lappihuan:
@oojacoboo i think type forcing as tested above should be fine, it would follow the current workflow of how to overwrite something between PHP types and GQL Types.

Lappihuan:
This would be supported with this commit:
https://github.com/thecodingmachine/graphqlite/pull/466/commits/1a21b830880cb68429e33e1c841392b624627bc6

oojacoboo:
Closing - assuming this may be resolved in version 6.0.  If not, please reopen.

PR #466, opened 2022-04-22
Bring input setters on par with type getters

Lappihuan:
The goal of this PR is to bring setters on Input Annotated Objects on par with getters.
Since the introduction of Input Types in #269 setters were very simple and did not bring any Field Middleware functionality that getters have (Autowire, Security, Right, Assert etc..).

With this Update setters have their own InputFieldMiddleware system that is pretty much 1:1 to the FieldMiddleware system allowing Autowire, Security and Right annotations to work.
Assertation would need to be added separately into the respective repositories for Symfony and Laravel.

While this seems like a relatively small addition, this allows for a very potent pattern with Doctrine Entities.
I will explain this pattern in depth as a example but this PR brings many more possibilities not outlined here.

Entities can now not only have computed fields on getters but also on setters leveraging the power of Autowire to set up references for example:
```php
#[ORM\Entity]
#[GQL\Type(default: true)]
#[GQL\Input(name: "UpdateAccountInput", update: true, default: true)]
class Account {
//...
    #[ORM\ManyToOne]
    #[ORM\JoinColumn(name: "lang_id", referencedColumnName: "lang_id", nullable: false)]
    private Language $lang;
//...
    #[GQL\Field]
    public function setLang(LanguageEnum $lang, #[GQL\Autowire] EntityManagerInterface $entityManager): void
    {
        $this->lang = $entityManager->getReference(Language::class, $lang->getValue());
    }

    #[GQL\Field]
    public function getLang(): LanguageEnum
    {
        return LanguageEnum::from($this->lang->getLangId());
    }
}
```
This allows for a very clean separation of ORM responsibility around the private properties and the public API using the setter and getter functions without the need to map between DTOs and Entities, there is even potential for nested graph traversing in mutations but this would need to be figured out.

Additionally using Doctrines [Embeddables](https://www.doctrine-project.org/projects/doctrine-orm/en/2.11/tutorials/embeddables.html#separating-concerns-using-embeddables) one can fully abstract the public data structure while still working with Doctrine Entities that can be persisted without any mapping.

Unrelated to this lib we even have a merge service that allows to persist partially hydrated Entities like this:
```php
    #[GQL\Mutation]
    public function updateAccount(
        ID $id,
        #[GQL\UseInputType(inputType: "UpdateAccountInput!")]
        Account $account
    ): Account
    {
        $account->setID($id);
        $merged = $this->mergeService->merge($account);
        $this->entityManager->flush();
        return $merge;
    }
```
The as for Validation directly on setters, i will make a PR on the [Symfony Validator Bridge](https://github.com/thecodingmachine/graphqlite-symfony-validator-bridge) Repo asap.

There is still documentation to be made around this feature and maybe some edge cases to be discussed.

@oojacoboo there are some commits left over from your input improvements since we initially based our branch on your.
Let me know if this needs cleanup or can be ignored.

Overall working on this PR gave us a lot of insights into how this lib is working and i already have a few more ideas on how to improve it :)

Looking forward to your feedback!


oojacoboo:
@Lappihuan I'll try to have a deeper look at this in the next few days.  

I'm not a huge fan of reusing the Doctrine entities as input types, personally.  I know other people are doing this as well and I don't see why we shouldn't allow more support around this.  There are a few things I really don't like, however, the autowiring being a big one.  Adding a service as a param to your domain model setters is really not ideal.  It works, but ideally your domain would be more pure and you'd use services outside of the domain models.  Autowiring is really a nasty hack overall, even outside the domain IMO.  That said, it's already implemented and I'm not sure the cat gets put in the bag around it.  Further, my more pure preferences may not be shared by many others.

I also still think support for setters is beneficial for DTO style Input types, allowing for modifications of input data before being "set" onto DTO properties.  Of course, these modifications could be done in operation controllers before being persisted or passed to domain models.  Regardless, there are situations where it's ideal to modify via a setter in the DTO.  We've been using constructors for this purpose since the Input type resolver instantiates the DTO, passing the requisite constructor arguments, prior to hydrating the DTO properties.  It'd be cleaner to handle some of this in setters and offers a location for some custom validation as well - that's the big win.

Lappihuan:
> Also, any thoughts on the Field annotation being used for input and output types? I'm a little concerned about the overall understanding on how these annotations work together with each other and the lack of clarity with their naming. In your example, you're going to get an input and output type with fields being mapped accordingly. However, to a noob, it's not inherently clear how this will happen.

This is exactly on of the edge cases. We are making the assumption that methods that start with `set` can only be InputFields and are filtered out for QueryFields.

While this works fine and seems to have no side effects it feels a tad bit hacky.

On the other hand i think PropertyAccessors like the one from Symfony have to make similar assumptions.
The one used in this lib does so too for example for getters:
https://github.com/thecodingmachine/graphqlite/blob/b69ba7e9d48fc91ef29504d93986bdb94d49d50a/src/Utils/PropertyAccessor.php#L27
I will check if i can reuse this.

oojacoboo:
@Lappihuan I'm wondering if we could enforce the use of the `for` option on the `Field` attribute if the definition is ambiguous.

Lappihuan:
@oojacoboo forcing `for` is not backwards compatible. but keep in mind, the field is not named `setLang` its named `lang` for input and output, its how getters worked since the beginning in this lib.
My concern was more with the technical implementation of it reather than the implicit separation, which should be natural to anyone familiar with graphqlite.
For a beginner, its a mather of having good docs explaining this very thing in a straight forward way.

Lappihuan:
> Could you please outline the changes and improvements here overall? It'd be nice to have a list of functionality changes.
Setters should now support everything getters do:
- parameter middlewares:
    - Autowire
    - InjectUser
    - Any user defined parameter middlewares
- field middlewares:
    - Security
    - Right
    - Logged
    - Support for Validation/Assertation is given but needs to be added to those packages
    - user defined field middlewares are not supported out of the box, but can easily be added by implementing the `InputFieldMiddlewareInterface`
- `inputType` property is now able to overwrite the `update` property as described in #371
- potential fix for crash with circular reference in input types #449, currently building a testcase to confirm.

oojacoboo:
@Lappihuan I've added a failing test, `@group PR-466`.  I came across the error while testing this PR on a rather large app.

Lappihuan:
@oojacoboo i fixed a minor issue where the inputFields property was was not initialized as empty array.
Was this the bug you are having?

oojacoboo:
@Lappihuan Yea, I guess the field annotations should be moved to the setters.  The issue is what’s being passed to the constructor isn’t correct.

Lappihuan:
@oojacoboo the field annotation is fine on properties, as long as there exists a public setter method.
I removed the exceptions from the setters and the test is not failing now.

not sure whats you see wrong with the constructor

oojacoboo:
Copied that over from another test, I wasn't able to test that it was properly failing.  Looks like it'll have to be adjusted.  We're getting the following error on our unit tests

```
TypeError: Foo\GraphQL\Operation\Employment\InputType\Employment::__construct(): Argument #1 ($startDate) must be of type ?Foo\Date, string given
```

`Foo\Date` is resolved with a custom root type mapper.

Lappihuan:
@oojacoboo constructor hydration proves to be challenging in the way it does not integrate with the whole middleware pattern.

The middlewares can go and wrap a field with other logic like for example Security, or any other user defined middleware. Now in those middlewares need to have access to the resolver and from those resolvers you are always supposed to get the DTO you are working with. In the case of the Security middleware you can use it to access properties on the DTO in the expression via `this` so if you have a `secret` Field on the DTO you can check it in the expression via `this.secret == "Foo"`. 

This works well in the case of setting properties directly or via setters since the DTO is already instantiated. But with Fields that are supposed be hydrated via constructor this whole system does not make sense. 🐔🥚

I feel like constructor hydration has to be a second class citizen and not support any middlewares.

I will try and come up with a implementation that determines if a field is hydrated via constructor and if so it will skip all the fancy FieldDescriptor and middleware stuff and just be used to instantiate the DTO before its passed to hydration via setters and properties.

Any toughs?

oojacoboo:
@Lappihuan I agree with skipping all the fancy business with the constructor.  Let's just get it instantiating and then we can do all the other middleware on the other fields being hydrated.  I'd be nice to check for fields being hydrated via the constructor to see if they're implementing any middleware logic, and if so, throw an exception.

There is one other option - fully constructing and hydrating the object and then post-processing all the middleware.  I suspect that this isn't worth it and would require recursively resolving all the fields in 2 passes - performance hit.

Lappihuan:
>There is one other option - fully constructing and hydrating the object and then post-processing all the middleware. I suspect that this isn't worth it and would require recursively resolving all the fields in 2 passes - performance hit.

Since i already do a second pass for setters and properties this wouldn't be a big deal.
But it assumes there is a way to do so. If the constructor is the only way to hydrate it would not work as expected.

I think is should be able to throw if middleware annotations are detected for constructor hydrated fields. :+1: 

Working on it right now.

oojacoboo:
@Lappihuan if you're already having to do 2 passes, using reflection, you can check which fields are being hydrated and simply re-hydrate them...  I don't love it, but it would retain functionality and I don't see any real issues with it.

Lappihuan:
@oojacoboo now i only do one pass for all constructor fields and one for setters/property fields. I detect early if the property is hydrated via constructor and if so check if middleware annotations are used if so i throw a exception explaining its not supported, else i just skip all the middleware stuff and resolve the properties for instantiation, after instantiation the new stuff kicks in. fields that are used in the constructor are pretty close to how they are handled in the lib right now.

oojacoboo:
@Lappihuan I'll give it a test again ASAP.

src/FieldsBuilder.php
```diff
@@ -824,4 +853,220 @@ private function getPrefetchMethodInfo(ReflectionClass $refClass, $reflector, ob
...
+                $inputFieldDescriptor = new InputFieldDescriptor();
+                $inputFieldDescriptor->setRefMethod($refMethod);
+                $inputFieldDescriptor->setIsUpdate($isUpdate);
+                $inputFieldDescriptor->setName($name);
+                $inputFieldDescriptor->setComment(trim($description));
+
+                $parameters = $refMethod->getParameters();
+                if ($injectSource === true) {
+                    $firstParameter = array_shift($parameters);
+                    // TODO: check that $first_parameter type is correct.
```
oojacoboo, line 899: Is this needed?  If not, let's update the code comment to something that's more helpful for other devs.

src/FieldsBuilder.php
```diff
@@ -342,8 +370,10 @@ private function getFieldsByMethodAnnotations($controller, ReflectionClass $refC
         $annotations = $this->annotationReader->getMethodAnnotations($refMethod, $annotationName);
         foreach ($annotations as $queryAnnotation) {
             $description = null;
+            $methodName = $refMethod->getName();
 
             if ($queryAnnotation instanceof Field) {
+                if (strpos($methodName, 'set') === 0) continue;
```
oojacoboo, line 376: It'd be good to stick with block statements

src/FieldsBuilder.php
```diff
@@ -824,4 +853,220 @@ private function getPrefetchMethodInfo(ReflectionClass $refClass, $reflector, ob
...
+            if ($fieldAnnotations instanceof Field) {
+                $for = $fieldAnnotations->getFor();
+                if ($typeName && $for && !in_array($typeName, $for)) {
+                    continue;
+                }
+                $description = $fieldAnnotations->getDescription();
+
+                $docBlockObj = $this->cachedDocBlockFactory->getDocBlock($refMethod);
+                $methodName = $refMethod->getName();
+                if (strpos($methodName, 'set') !== 0) continue;
```
oojacoboo, line 884: It'd be good to stick with block statements

src/FieldsBuilder.php
```diff
@@ -824,4 +853,220 @@ private function getPrefetchMethodInfo(ReflectionClass $refClass, $reflector, ob
...
+                $inputFieldDescriptor->setMiddlewareAnnotations($this->annotationReader->getMiddlewareAnnotations($refMethod));
+
+                $field = $this->inputFieldMiddleware->process($inputFieldDescriptor, new class implements InputFieldHandlerInterface {
+                    public function handle(InputFieldDescriptor $inputFieldDescriptor): ?InputField
+                    {
+                        return InputField::fromFieldDescriptor($inputFieldDescriptor);
+                    }
+                });
+
+
```
oojacoboo, line 941: Unnecessary blank lines

src/FieldsBuilder.php
```diff
@@ -824,4 +853,220 @@ private function getPrefetchMethodInfo(ReflectionClass $refClass, $reflector, ob
...
+            if ($annotation instanceof Field) {
+                $for = $annotation->getFor();
+                if ($typeName && $for && !in_array($typeName, $for)) {
+                    continue;
+                }
+
+                $description = $annotation->getDescription();
+
+
+                $name = $annotation->getName() ?: $refProperty->getName();
```
oojacoboo, line 984: Wouldn't the null coalescing operator be more appropriate here?  `??`

src/FieldsBuilder.php
```diff
@@ -824,4 +853,220 @@ private function getPrefetchMethodInfo(ReflectionClass $refClass, $reflector, ob
...
+
+                $inputType = $annotation->getInputType();
+                $constructerParameters = $this->getClassConstructParameterNames($refClass);
+                $inputProperty = $this->typeMapper->mapInputProperty($refProperty, $docBlock, $name, $inputType, $defaultProperties[$refProperty->getName()] ?? null, $isUpdate ? true : null);
+
+                if (! $description) {
+                    $description = $inputProperty->getDescription();
+                }
+
+                if(in_array($name, $constructerParameters)) {
```
oojacoboo, line 994: Space after `if`... `if (`

src/FieldsBuilder.php
```diff
@@ -824,4 +853,220 @@ private function getPrefetchMethodInfo(ReflectionClass $refClass, $reflector, ob
...
+                    $middlewareAnnotations = $this->annotationReader->getPropertyAnnotations($refProperty, MiddlewareAnnotationInterface::class);
+                    if($middlewareAnnotations !== []){
```
oojacoboo, line 996: Space after `if`.  Not sure why phpcs isn't picking up on this.

src/InputField.php
```diff
@@ -0,0 +1,215 @@
...
+            'name' => $name,
+            'type' => $type,
+            'description' => $comment
+        ];
+
+        if (!(!$hasDefaultValue || $isUpdate)) {
+            $config['defaultValue'] = $defaultValue;
+        }
+
+        if($originalResolver !== null && $resolver !== null){
```
oojacoboo, line 58: Space after `if`

src/InputField.php
```diff
@@ -0,0 +1,215 @@
...
+            $fieldDescriptor->isUpdate(),
+            $fieldDescriptor->hasDefaultValue(),
+            $fieldDescriptor->getDefaultValue(),
+        );
+    }
+
+    public static function fromFieldDescriptor(InputFieldDescriptor $fieldDescriptor): self
+    {
+        $arguments = $fieldDescriptor->getParameters();
+//        if ($fieldDescriptor->getPrefetchMethodName() !== null) {
```
oojacoboo, line 178: Unused code should be removed

src/InputFieldDescriptor.php
```diff
@@ -0,0 +1,345 @@
...
+
+    /**
+     * @param mixed|null $defaultValue
+     */
+    public function setDefaultValue($defaultValue): void
+    {
+        $this->defaultValue = $defaultValue;
+    }
+
+
```
oojacoboo, line 123: Extra lines

src/InputFieldDescriptor.php
```diff
@@ -0,0 +1,345 @@
...
+     */
+    public function setCallable(callable $callable): void
+    {
+        if ($this->originalResolver !== null) {
+            throw new GraphQLRuntimeException('You cannot modify the callable via setCallable because it was already used. You can still wrap the callable using getResolver/setResolver');
+        }
+        $this->callable = $callable;
+        $this->targetMethodOnSource = null;
+        $this->targetPropertyOnSource = null;
+//        $this->magicProperty = null;
```
oojacoboo, line 206: Remove unused code

src/InputFieldDescriptor.php
```diff
@@ -0,0 +1,345 @@
...
+
+    public function setTargetMethodOnSource(string $targetMethodOnSource): void
+    {
+        if ($this->originalResolver !== null) {
+            throw new GraphQLRuntimeException('You cannot modify the target method via setTargetMethodOnSource because it was already used. You can still wrap the callable using getResolver/setResolver');
+        }
+        $this->callable = null;
+        $this->targetMethodOnSource = $targetMethodOnSource;
+        $this->targetPropertyOnSource = null;
+//        $this->magicProperty = null;
```
oojacoboo, line 217: Remove unused code

src/InputFieldDescriptor.php
```diff
@@ -0,0 +1,345 @@
...
+
+        if (is_callable($this->callable)) {
+            /** @var callable&array{0:object, 1:string} $callable */
+            $callable = $this->callable;
+            $this->originalResolver = new ServiceResolver($callable);
+        } elseif ($this->targetMethodOnSource !== null) {
+            $this->originalResolver = new SourceResolver($this->targetMethodOnSource);
+        } elseif ($this->targetPropertyOnSource !== null) {
+            $this->originalResolver = new SourceInputPropertyResolver($this->targetPropertyOnSource);
+//        } elseif ($this->magicProperty !== null) {
```
oojacoboo, line 319: Unused code

src/Middlewares/AuthorizationInputFieldMiddleware.php
```diff
@@ -0,0 +1,91 @@
...
+    public function process(InputFieldDescriptor $inputFieldDescriptor, InputFieldHandlerInterface $inputFieldHandler): ?InputField
+    {
+        $annotations = $inputFieldDescriptor->getMiddlewareAnnotations();
+
+        $loggedAnnotation = $annotations->getAnnotationByType(Logged::class);
+        assert($loggedAnnotation === null || $loggedAnnotation instanceof Logged);
+        $rightAnnotation = $annotations->getAnnotationByType(Right::class);
+        assert($rightAnnotation === null || $rightAnnotation instanceof Right);
+
+//        $failWith = $annotations->getAnnotationByType(FailWith::class);
```
oojacoboo, line 47: Unused code

src/Middlewares/AuthorizationInputFieldMiddleware.php
```diff
@@ -0,0 +1,91 @@
...
+//            Assert::isInstanceOf($type, OutputType::class);
+//            $inputFieldDescriptor->setType($type);
+//        }
+        if ($this->isAuthorized($loggedAnnotation, $rightAnnotation)) {
+            return $inputFieldHandler->handle($inputFieldDescriptor);
+        }
+
+        $hideIfUnauthorized = $annotations->getAnnotationByType(HideIfUnauthorized::class);
+        assert($hideIfUnauthorized instanceof HideIfUnauthorized || $hideIfUnauthorized === null);
+
```
oojacoboo, line 63: More unused code

src/Middlewares/InputFieldHandlerInterface.php
```diff
@@ -0,0 +1,21 @@
...
+
+declare(strict_types=1);
+
+namespace TheCodingMachine\GraphQLite\Middlewares;
+
+use TheCodingMachine\GraphQLite\InputField;
+use TheCodingMachine\GraphQLite\InputFieldDescriptor;
+
+/**
+ * @unstable See https://graphqlite.thecodingmachine.io/docs/semver.html
```
oojacoboo, line 11: What is meant by this annotation?  Link is broken.

src/Middlewares/InputFieldMiddlewareInterface.php
```diff
@@ -0,0 +1,18 @@
...
+
+namespace TheCodingMachine\GraphQLite\Middlewares;
+
+use TheCodingMachine\GraphQLite\InputField;
+use TheCodingMachine\GraphQLite\InputFieldDescriptor;
+
+/**
+ * A middleware use to process annotations when evaluating a input field
+ *
+ * @unstable See https://graphqlite.thecodingmachine.io/docs/semver.html
```
oojacoboo, line 13: What is meant by this annotation? Link is broken.

src/Middlewares/SecurityInputFieldMiddleware.php
```diff
@@ -0,0 +1,144 @@
...
+    /*private $logger;*/
+    /** @var AuthorizationServiceInterface */
+    private $authorizationService;
+
+    public function __construct(ExpressionLanguage $language, AuthenticationServiceInterface $authenticationService, AuthorizationServiceInterface $authorizationService/*, ?LoggerInterface $logger = null*/)
+    {
+        $this->language = $language;
+        $this->authenticationService = $authenticationService;
+        $this->authorizationService = $authorizationService;
+        /*$this->logger = $logger;*/
```
oojacoboo, line 41: Unused code

src/Middlewares/SecurityInputFieldMiddleware.php
```diff
@@ -0,0 +1,144 @@
...
+    {
+        $annotations = $inputFieldDescriptor->getMiddlewareAnnotations();
+        /** @var Security[] $securityAnnotations */
+        $securityAnnotations = $annotations->getAnnotationsByType(Security::class);
+
+        if (empty($securityAnnotations)) {
+            return $inputFieldHandler->handle($inputFieldDescriptor);
+        }
+
+//        $failWith = $annotations->getAnnotationByType(FailWith::class);
```
oojacoboo, line 54: Unused code

src/Middlewares/SecurityInputFieldMiddleware.php
```diff
@@ -0,0 +1,144 @@
...
+            'authorizationService' => $this->authorizationService, // Used by the is_granted expression language function.
+            'authenticationService' => $this->authenticationService, // Used by the is_logged expression language function.
+            'this' => $callable->getObject(),
+        ];
+
+        $argsName = array_keys($parameters);
+        $argsByName = array_combine($argsName, $args);
+        Assert::isArray($argsByName);
+
+        /*if ($diff = array_intersect(array_keys($variables), array_keys($argsName))) {
```
oojacoboo, line 125: Unused code

src/Types/InputType.php
```diff
@@ -45,32 +45,25 @@ public function __construct(
...
-            foreach ($this->fields as $name => $field) {
-                $type = $field->getType();
+        $fields = function () use ($isUpdate, $inputName, $className, $fieldsBuilder) {
+            $inputFields = $fieldsBuilder->getInputFields($className, $inputName, $isUpdate);
 
-                if ($isUpdate && $type instanceof NonNull) {
-                    $type = $type->getWrappedType();
+            $fieldConfigs = [];
+            foreach($inputFields as $field){
+                if($field->forConstructorHydration()){
```
oojacoboo, line 54: Space after `if`

src/Types/InputType.php
```diff
@@ -45,32 +45,25 @@ public function __construct(
...
-            $fields = [];
-            foreach ($this->fields as $name => $field) {
-                $type = $field->getType();
+        $fields = function () use ($isUpdate, $inputName, $className, $fieldsBuilder) {
+            $inputFields = $fieldsBuilder->getInputFields($className, $inputName, $isUpdate);
 
-                if ($isUpdate && $type instanceof NonNull) {
-                    $type = $type->getWrappedType();
+            $fieldConfigs = [];
+            foreach($inputFields as $field){
```
oojacoboo, line 53: Space after `foreach`

src/Types/InputType.php
```diff
@@ -88,27 +81,30 @@ public function __construct(
...
     {
-        $mappedValues = [];
-        foreach ($this->fields as $field) {
-            $name = $field->getName();
-            if (! array_key_exists($name, $args)) {
+        $constructorArgs = [];
+        foreach ($this->constructorInputFields as $constructorInputField) {
+            $name = $constructorInputField->name;
+            $resolve = $constructorInputField->getResolve();
+            if(!array_key_exists($name, $args)) {
```
oojacoboo, line 88: Space after `if`

tests/Fixtures/CircularInputReference/Controllers/CircularController.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
```
oojacoboo, line 2: Superfluous lines

tests/Fixtures/CircularInputReference/Controllers/CircularController.php
```diff
@@ -0,0 +1,20 @@
...
+
+namespace TheCodingMachine\GraphQLite\Fixtures\CircularInputReference\Controllers;
+
+
```
oojacoboo, line 6: Superfluous lines

tests/Fixtures/CircularInputReference/Types/CircularInputA.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+class CircularInputA
+{
+
+    /**
+     * @Field(inputType="CircularInputBInput")
+     * @var CircularInputB
+     */
+    private $circularInputB;
+
```
oojacoboo, line 19: Superfluous lines

tests/Fixtures/CircularInputReference/Types/CircularInputB.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+class CircularInputB
+{
+
+    /**
+     * @Field(inputType="CircularInputAInput")
+     * @var CircularInputA
+     */
+    private $circularInputA;
+
```
oojacoboo, line 19: Superfluous lines

oojacoboo (changes requested):
@Lappihuan good news, I've had a chance to test the latest code with this PR on a large application - all tests are passing!

I've added a code review here.  We probably should focus on getting `phpcs` rules updated to catch a lot of these code style comments I've made, then get that passing.

Looking forward to getting this merged in.  Thanks again for all the work on this!

src/FieldsBuilder.php
Lappihuan: this is copied from the getter part of the fields builder which has this comment, from what i understand, there should be a sanity check if the injected source is actually the same Type as hinted.

I'd leave it for now.

src/FieldsBuilder.php
Lappihuan: its also just copied from the equivalent code for output types, might have been a backwards compatible reason for it.

i think it should probably be updated in the whole library at some point, maybe using rector to automate it.

src/InputFieldDescriptor.php
Lappihuan: the magic property code is to be enabled in a future PR

src/InputFieldDescriptor.php
Lappihuan: the magic property code is to be enabled in a future PR

src/InputFieldDescriptor.php
```diff
@@ -0,0 +1,345 @@
...
+
+    public function setTargetPropertyOnSource(?string $targetPropertyOnSource): void
+    {
+        if ($this->originalResolver !== null) {
+            throw new GraphQLRuntimeException('You cannot modify the target method via setTargetMethodOnSource because it was already used. You can still wrap the callable using getResolver/setResolver');
+        }
+        $this->callable = null;
+        $this->targetMethodOnSource = null;
+        $this->targetPropertyOnSource = $targetPropertyOnSource;
+//        $this->magicProperty = null;
```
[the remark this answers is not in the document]
Lappihuan: the magic property code is to be enabled in a future PR

src/InputFieldDescriptor.php
Lappihuan: the magic property code is to be enabled in a future PR

src/Middlewares/InputFieldHandlerInterface.php
Lappihuan: copied from FieldHandlerInterface

src/Middlewares/InputFieldMiddlewareInterface.php
Lappihuan: copied from FieldMiddlewareInterface

src/Middlewares/SecurityInputFieldMiddleware.php
```diff
@@ -0,0 +1,144 @@
...
+    /** @var ExpressionLanguage */
+    private $language;
+    /** @var AuthenticationServiceInterface */
+    private $authenticationService;
+    /** @var LoggerInterface|null */
+    /*private $logger;*/
+    /** @var AuthorizationServiceInterface */
+    private $authorizationService;
+
+    public function __construct(ExpressionLanguage $language, AuthenticationServiceInterface $authenticationService, AuthorizationServiceInterface $authorizationService/*, ?LoggerInterface $logger = null*/)
```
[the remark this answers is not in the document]
Lappihuan: this is copied from the original SecurityFieldMiddleware which has a todo to add Logger support.
i assume this is preparation for that

src/Middlewares/SecurityInputFieldMiddleware.php
Lappihuan: connected to the logger stuff above

Lappihuan:
@oojacoboo let me know if there is something to change on the unresolved points

src/InputFieldDescriptor.php
oojacoboo: Let's add a comment with any commented out code.  In general, we should remove virtually all commented out code, but small lines like this are fine if we can share with others why it's there.

src/InputFieldDescriptor.php
oojacoboo: Let's add a comment with any commented out code. In general, we should remove virtually all commented out code, but small lines like this are fine if we can share with others why it's there.

src/InputFieldDescriptor.php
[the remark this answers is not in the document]
oojacoboo: Let's add a comment with any commented out code. In general, we should remove virtually all commented out code, but small lines like this are fine if we can share with others why it's there.

src/InputFieldDescriptor.php
oojacoboo: Let's add a comment with any commented out code. In general, we should remove virtually all commented out code, but small lines like this are fine if we can share with others why it's there.

src/Middlewares/InputFieldHandlerInterface.php
oojacoboo: Let's remove it then.  The link is dead and doesn't make sense.

src/Middlewares/InputFieldMiddlewareInterface.php
oojacoboo: Let's remove it then. The link is dead and doesn't make sense.

src/Middlewares/SecurityInputFieldMiddleware.php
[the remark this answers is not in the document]
oojacoboo: Let's add a comment for the reasoning for it being there.

src/Middlewares/SecurityInputFieldMiddleware.php
oojacoboo: Let's add a comment for the reasoning for it being there.

oojacoboo:
@Lappihuan thanks, added some more comments.  In general, if we're leaving commented out code, we should at least have a comment saying why it's there so it doesn't cause any confusion.

oojacoboo (approved):

oojacoboo:
@Lappihuan I've merged in this PR.  Documentation wasn't completed and I'm doing that now, in addition to versioning the docs for a 6.0 release.

PR #467, opened 2022-04-24
Support native union return typing

### LINKED ISSUE #462: PHP 8 Union type fails while union type in doc block works
I get this error when using the output type:

```php
#[Field]
public function origin(Foo $foo): Bar|Baz
{
  return $foo->origin;
}
```

```
In TypeHandler.php line 367:

  assert($type instanceof ReflectionNamedType)
```

The output types for Bar and Baz are already defined.

Adding the union type to the docblock works:

```php
/**
* @return Bar|Baz
*/
#[Field]
public function origin(Foo $foo)
{
  return $foo->origin;
}
```

I assume this is a missing php 8 features

aszenz:
Fixes #462

### DIFF
--- a/src/Mappers/Parameters/TypeHandler.php
+++ b/src/Mappers/Parameters/TypeHandler.php
@@ -27,6 +27,7 @@ use ReflectionNamedType;
 use ReflectionParameter;
 use ReflectionProperty;
 use ReflectionType;
+use ReflectionUnionType;
 use TheCodingMachine\GraphQLite\Annotations\HideParameter;
 use TheCodingMachine\GraphQLite\Annotations\ParameterAnnotations;
 use TheCodingMachine\GraphQLite\Annotations\UseInputType;
@@ -42,6 +43,7 @@ use TheCodingMachine\GraphQLite\Types\ArgumentResolver;
 use TheCodingMachine\GraphQLite\Types\TypeResolver;
 use Webmozart\Assert\Assert;
 
+use function array_map;
 use function array_merge;
 use function array_unique;
 use function assert;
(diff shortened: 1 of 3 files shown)


tests/Mappers/Parameters/TypeMapperTest.php
```diff
@@ -77,6 +101,10 @@ private function dummy() {
 
     }
 
+    private function objectUnion(): TestObject|TestObject2 {
```
aszenz, line 104: I also tried `null|TestObject|TestObject2` but that failed, not sure if the library needs to support that

aszenz:
@oojacoboo Any idea how to fix the test for php 7.4

oojacoboo:
@aszenz PHP didn't support union types until 8.0.  The parser for 7.4 doesn't understand the syntax.  Makes sure the tests are only running for 8.0.  You can look at how `tests/Fixtures81` is implemented.  If necessary, create a new `Fixtures80` directory.

Tests are run through `.github/workflows/continuous_integration.yml`.

aszenz:
> @aszenz PHP didn't support union types until 8.0. The parser for 7.4 doesn't understand the syntax. Make sure the tests are only running for 8.0. You can look at how `tests/Fixtures81` is implemented. If necessary, create a new `Fixtures80` directory.
> 
> Tests are run through `.github/workflows/continuous_integration.yml`

Done :+1:

src/Mappers/Parameters/TypeHandler.php
```diff
@@ -408,17 +410,36 @@ private function appendTypes(Type $type, ?Type $docBlockType): Type
...
+                $phpdocType = new Nullable($phpdocType);
+            }
 
-        return $phpdocType;
+            return $phpdocType;
+        }
+        return new Compound(
+            array_map(
+                function ($namedType) use ($reflectionClass): Type {
+                    \assert($namedType instanceof ReflectionNamedType);
```
aszenz, line 429: had to add this assert to satisfy phpstan for php7.4

oojacoboo:
@aszenz awesome thanks and for the PR!

Issue #436, opened 2022-01-20
Support Generics style PHPDoc @return typing


oojacoboo:
When returning a Collection, you can annotate the `@return` type like:

```php
/**
 * @return Article[]
 */
#[GraphQLite\Field]
public function getArticles(): Collection; 
```

However, this doesn't tell many static analyzers that the return type is a Collection.  Maybe some smarter analyzers will check the actual return type, if typed, and use both combined.  I'm unsure about that.  But, there is value in using the Generics syntax for those that don't.

```php
/**
 * @return Collection<Article>
 */
#[GraphQLite\Field]
public function getArticles(): Collection;
```

This syntax is quickly becoming somewhat of a standard at this point.  It's supported by Psalm, Phan, PHPStan and I'm sure others.  It was [implemented by PHPStorm](https://blog.jetbrains.com/phpstorm/2021/07/phpstorm-2021-2-beta/).

Many of you are probably familiar with this syntax and probably use it.  For compatibility reasons, we should support both of these syntaxes when evaluating an `@return` annotation for the sake of generating GraphQL Types.  Currently, an error is thrown as it tries to create a Type definition out of `Collection`

Lappihuan:
wouldn't #321 fix this?

oojacoboo:
@Lappihuan yes, probably so - we need tests there to confirm.

oojacoboo:
Generics support was added with #468 and released with version 6.0

PR #468, opened 2022-04-24
Support generics typed iterables for output types

andrew-demb:
refs: #436 

Unfortunately, I'm new to the GraphQLite codebase, so it is hard for me to find a clean implementation (and to know about possible drawbacks of chosen way to handle generic iterables)


oojacoboo:
Hey @andrew-demb.  Thanks for this.  Did you see #321?

oojacoboo:
Hey @andrew-demb.  Thanks for this.  Did you see #321?

andrew-demb:
Hey @oojacoboo .
Yes, I saw. 

That PR doesn't solve the problem.
IteratorTypeMapper only handle union types (`Compound`), so provided fix is not enough.
You can check my attempt (which passes the test with generic phpdoc) to solve the issue with IteratorTypeMapper in the first commit.

oojacoboo:
Thanks for clarifying @andrew-demb.  I think it'd be good to also have a test for the full schema parser, maybe something in `tests/Integration/EndToEndTest.php` or even a new test class in `tests/Integration`.

andrew-demb:
@oojacoboo I found a fixture controller with different typed methods and added to it new methods with generic types.
Adding new methods bubbles up another problem with merging phpdoc+return type types.
So I added a patch to this merging process to be more accurate (it may look a bit hacky, but I have no other ideas for now).

oojacoboo:
Does it make more sense to handle the generic types in this condition?

```php
if ($innerType instanceof Array_ || $innerType instanceof Iterable_ || $innerType instanceof Mixed_) {~
```

... instead of the `else`?  It seems that if condition is attempting to handle iterable types.  Are there types that need generic style typing that aren't an iterable?

andrew-demb:
> Does it make more sense to handle the generic types in this condition in the `TypeHandler`?
> 
> ```
> if ($innerType instanceof Array_ || $innerType instanceof Iterable_ || $innerType instanceof Mixed_) {~
> ```
> 
> ... instead of the `else`? It seems that if condition is attempting to handle iterable types. Are there types that need generic style typing that aren't an iterable?

Seems like yes, it will be a more appropriate place.
I simply moved the check to the mentioned condition.

It is a bit clumsy (since inside condition logic will check dock block type for throwing new exception), but another way may require a check for implementation of `\Iterator | \IteratorAggregate` (via reflection).

src/Mappers/Parameters/TypeHandler.php
```diff
@@ -338,7 +339,17 @@ private function mapType(
...
+            // Try to match generic phpdoc-provided iterables with non-generic return-type-provided iterables
+            // Example: (return type `\ArrayObject`, phpdoc `\ArrayObject<string, TestObject>`)
+            || ($innerType instanceof Object_
+                && $docBlockType instanceof Collection
+                && (string)$innerType->getFqsen() === (string)$docBlockType->getFqsen()
+            )
+        ) {
             // We need to use the docBlockType
             if ($docBlockType === null) {
                 throw CannotMapTypeException::createForMissingPhpDoc($innerType, $reflector, $argumentName);
```
andrew-demb, line 355: phpstan complies about the unexpected type (`Object_`) for `$type` argument.
Fixing this may require to find a wording for exception message.


```
Parameter #1 $type of static method TheCodingMachine\GraphQLite\Mappers\CannotMapTypeException::createForMissingPhpDoc() expects phpDocumentor\Reflection\Types\Array_|phpDocumentor\Reflection\Types\Iterable_|phpDocumentor\Reflection\Types\Mixed_, phpDocumentor\Reflection\Types\Array_|phpDocumentor\Reflection\Types\Iterable_|phpDocumentor\Reflection\Types\Mixed_|phpDocumentor\Reflection\Types\Object_ given.
```


I need suggestions about:
- is it an OK place for the condition?
- should I change the condition to another one?

Thanks

oojacoboo:
@andrew-demb I'm not sure why I can't reply on the thread with your question.  But, as for your question, I'm not sure I understand this error `phpDocumentor\Reflection\Type\Object_` is an instance of `phpDocumentor\Reflection\Type`, which the `createForMissingPhpDoc` method argument, `$type`, is typed.

andrew-demb:
@oojacoboo phpdoc above method narrow down possible type for input parameter
https://github.com/thecodingmachine/graphqlite/blob/53d001897eb3c522e07c5a4a5ce8109490ef73cb/src/Mappers/CannotMapTypeException.php#L135

And method contain specific handing for every type (to generate proper exception message)

oojacoboo:
@andrew-demb I see no reason not to expand the phpdoc annotation to support this case.

andrew-demb:
> @andrew-demb I see no reason not to expand the phpdoc annotation to support this case.

The problem was with:
- what exception message should be shown for the developer in that case
- this exception should never happen, since the condition contains a check for not-null doc block type (and specific dock block implementation - `Collection`)

Since it is not necessary, I didn't change the exception message and just allows to bypass the `Object_` instance.

oojacoboo:
@andrew-demb I see - thanks.  In a case like this, I'd just add support and pretend that it could be reached.  Future changes could cause that to be the case.  Additionally, this method may be reused in other locations.  It's best to write a function to handle the logic of its inputs, irregardless of current call stack flows.

andrew-demb:
Thanks for suggestions @oojacoboo 

What can I do to proceed the PR?

oojacoboo:
What happens with the following generic annotations?

```php
use Acme\Foo;

@return iterable<int, Foo>
@return \Iterator<string, int>
@return \IteratorAggregate<int, bool>
@return \Traversable<string, Foo>
```

andrew-demb:
They will be all converted to ListOfType with proper value type.
Unfortunately information about string keys won't be preserved (I didn't find same functionality for IteratorTypeMapper too).

I will add unit tests for these cases to track behavior.

oojacoboo:
This looks great @andrew-demb.  Thanks again for the PR!

Issue #440, opened 2022-02-11
Mutation namespaces


ChvilaMartin:
Hello, hopefully this is not duplication of existing issue.

Is there support for Mutation namespaces like described there: https://graphql-rules.com/rules/mutation-namespaces?

What I want is to be able to have something like this

```
UserMutations {
   create(name: String!): String!,
   update(id: String!, name: String!): Boolean!,
   ....
},
PostMutations {
   ...
}
```

oojacoboo:
@ChvilaMartin that's pretty cool.  Unfortunately, that's not currently supported.  I also don't see any support for this with https://github.com/webonyx/graphql-php.  We'd need support there before adding support for that here.

enumag:
@oojacoboo Webonyx supports it just fine. Or rather no additional support on their sice is needed to implement it. I'm using it on another webonyx-based project. Here is how it works:

First I have types like these, representing the mutation namespaces:

```php
final class UserMutationsType extends ObjectType
{
    public function __construct()
    {
        parent::__construct([
            'name' => 'UserMutations',
            'description' => null,
            'fields' => function () {
                return [
                    'create' => [
                        // definition of the create mutation
                    ],
                ];
            },
            'interfaces' => [],
            'isTypeOf' => null,
            'resolveField' => null,
        ]);
    }
}
```

Next the MutationType needs to use these:

```php
final class MutationType extends ObjectType
{
    public function __construct(ContainerInterface $typeResolver)
    {
        parent::__construct([
            'name' => 'Mutation',
            'description' => null,
            'fields' => function () use ($typeResolver) {
                return [
                    'user' => [
                        'type' => $typeResolver->get('UserMutations'),
                        'args' => [],
                        'resolve' => function ($value, $args, $context, ResolveInfo $info) {
                            return [];
                        },
                        'description' => null,
                        'deprecationReason' => null,
                        'complexity' => null,
                    ],
                ];
            },
            'interfaces' => [],
            'isTypeOf' => null,
            'resolveField' => null,
        ]);
    }
}
```

Easy enough in my opinion.

oojacoboo:
@enumag thanks for the clarification here.  This seems like a worthwhile improvement if someone wants to add a PR.

Lappihuan:
I've been thinking about this lately and feel like a Namespace Attribute on Controllers would be the most logical.
Imagine:
```php
#[GQL\Namespace]
class UserController {
    #[GQL\Query]
    public function user()...

[...]

PR #475, opened 2022-04-27
Muatation namespace

Lappihuan:
This PR adds the ability to turn a `Controller` into a Mutation Namespace as described in #440

The specific Implementation adds a `MutationNamespace` Attribute/Annotation which can optionally be added to a Controller to wrap all its mutations into a Mutation Namespace.

This would look something like this:
```php
#[GQL\MutationNamespace]
class UserController {
    #[GQL\Query]
    public function user()...

    #[GQL\Mutation]
    public function create()...

    #[GQL\Mutation]
    public function update()...

    #[GQL\Mutation]
    public function delete()...
```
Which turns into:
```sdl
UserMutations {
   create()...
   update()....
   delete()...
}
```
There is a optional `name` parameter on the `MutationNamespace` Attribute to rename automatically generated Namespace.
The automatically generated Namespace is the `Classname` - `Controller` + `Muatations` which turns UserController into UserMutations.

All queries in the Controller are completely unaffected.

### DIFF
--- a/src/Annotations/MuatationNamespace.php
+++ b/src/Annotations/MuatationNamespace.php
@@ -1,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TheCodingMachine\GraphQLite\Annotations;
+
+use Attribute;
+
+/**
+ * @Annotation
+ * @Target({"CLASS"})
+ * @Attributes({
+ *   @Attribute("name", type = "string"),
+ * })
+ */
+#[Attribute(Attribute::TARGET_CLASS)]
+class MuatationNamespace
+{
+
+    /** @var string|null */
+    private $name;
+
+    /**
+     * @param mixed[] $attributes
+     */
+    public function __construct(array $attributes = [], ?string $name = null)
+    {
+        $this->name = $name ?? $attributes['name'] ?? null;
+    }
+
+    /**
+     * Returns the GraphQL name for this type.
+     */
+    public function getName(): ?string
+    {
+        return $this->name;
+    }
+}
--- a/src/AnnotationReader.php
+++ b/src/AnnotationReader.php
@@ -22,6 +22,7 @@ use TheCodingMachine\GraphQLite\Annotations\Factory;
 use TheCodingMachine\GraphQLite\Annotations\Input;
 use TheCodingMachine\GraphQLite\Annotations\MiddlewareAnnotationInterface;
 use TheCodingMachine\GraphQLite\Annotations\MiddlewareAnnotations;
+use TheCodingMachine\GraphQLite\Annotations\MuatationNamespace;
 use TheCodingMachine\GraphQLite\Annotations\ParameterAnnotationInterface;
 use TheCodingMachine\GraphQLite\Annotations\ParameterAnnotations;
 use TheCodingMachine\GraphQLite\Annotations\SourceFieldInterface;
@@ -277,6 +278,16 @@ class AnnotationReader
         }
 
         return $type;
+    }
+
+    /**
+     * @param ReflectionClass<T> $refClass
+     *
+     * @template T of object
+     */
+    public function getNamespaceAnnotation(ReflectionClass $refClass): ?MuatationNamespace
+    {
+        return $this->getClassAnnotation($refClass, MuatationNamespace::class);
     }
 
     /**
--- a/src/FieldsBuilder.php
+++ b/src/FieldsBuilder.php
@@ -121,17 +121,27 @@ class FieldsBuilder
      */
     public function getMutations(object $controller): array
     {
-//        return $this->getFieldsByAnnotations($controller, Mutation::class, false);
         $refClass = new ReflectionClass($controller);
-        $namespace = $this->annotationReader->getNamespaceAnnotation($refClass);
-        if(!$namespace){
+        $namespaceAnnotation = $this->annotationReader->getNamespaceAnnotation($refClass);
+        if(!$namespaceAnnotation){
             return $this->getFieldsByAnnotations($controller, Mutation::class, false);
-        } else {
-            dump($namespace);
-            $muatations = $this->getFieldsByAnnotations($controller, Mutation::class, false);
-            dump($muatations);
-            return $muatations;
         }
+
+        $namespaceFieldName = str_replace("Controller", "", $refClass->getShortName());
+        $namespaceName = $namespaceAnnotation->getName() ?? $namespaceFieldName."Mutations";
+
+        $namespaceType = new MutableObjectType([
+            'name' => $namespaceName,
+            'fields' => function () use ($controller) {
+                return $this->getFieldsByAnnotations($controller, Mutation::class, false);
+            }
+        ]);
+        $namespaceType->freeze();
+        $namespaceField = FieldDefinition::create([
+            'name' => strtolower($namespaceFieldName),
+            'type' => $namespaceType
+        ]);
+        return [$namespaceFieldName => $namespaceField];
     }
 
     /**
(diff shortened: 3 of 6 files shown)


oojacoboo:
Thanks for this @Lappihuan.

Firstly, is there any reason why we're limiting namespaces to mutations and not queries as well?  I'm assuming we can also namespace queries?  And, in this case, a `GQL\Namespace` attribute seems appropriate.  Can types be namespaced?  Should it be `GQL\OperationNamespace` if not?

Regarding the default namespacing, this is going to be contentious, as everyone is going to have different "controller" class names.  For instance, our "controller" namespace is `Operation/Foo/Mutation`.  I don't think we can settle on a default naming scheme that will have much unanimity.  At any rate, I don't think the chosen default is ideal.  We should probably just stick with whatever the class is, or even the entire namespace of the class.  I suspect most people will need to define the `name` option.

src/Annotations/MuatationNamespace.php
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * @Annotation
+ * @Target({"CLASS"})
+ * @Attributes({
+ *   @Attribute("name", type = "string"),
+ * })
+ */
+#[Attribute(Attribute::TARGET_CLASS)]
+class MuatationNamespace
```
andrew-demb, line 17:
```suggestion
class MutationNamespace
```

Lappihuan:
>Firstly, is there any reason why we're limiting namespaces to mutations and not queries as well? I'm assuming we can also namespace queries? And, in this case, a GQL\Namespace attribute seems appropriate. Can types be namespaced? Should it be GQL\OperationNamespace if not?

@oojacoboo 
This is a simple convention to group C*UD from CRUD since GraphQL only offers two Root queries (Mutation, Query) and since R from CRUD already maps 1:1 to Query, CUD could be grouped into a "Namespace".
But all thats it, is a parent Mutation `user` with the subType `UserMutations` which has all the mutations as fields.

As for the Attribute name, `Namespace` is a reserved word in PHP thats why i went with `MutationNamespace` since it describes the feature best.

I don't see this being usefull for Queries. And about Types, its just that a Type with fields, you'd be better to nest two types.

As for the naming, its just masking `Controller` out and Adds `Mutations` i don't mind adding a bunch common words to mask out and as soon as that convention doesn't work the `name` property can be used.

I actually forgot to set this as a Draft since there is quite a bit of stuff in the Technical implementation that should be improved.
Or at least discussed.

I'm unsure if those fields and Types should be created by a `TypeMapper` so they can be cached and so on. And if so i don't know way of doing so, but it might be easier than i think.

andrew-demb:
Can we add a possibility to specify a namespace as of `#[Query] / #[Mutation]` attributes too?
Or allow using new separated namespace attribute specified over methods?

We use invokable controllers with a single method in every class, so adding attribute over class (operation-related attribute already specified over method)  would be a bit uncomfortable :)

src/Annotations/MuatationNamespace.php
Lappihuan: Thanks, for some reason i keep writing `Muatation` :smile:

Lappihuan:
@andrew-demb do you have an example of such a invokable Controller?
I'm wondering if you shouldn't just use `#[Mutation]` as your "namespace" level and use `#[Field]` for whatever is nested in it.
Which would already be supported for your case.
See https://github.com/thecodingmachine/graphqlite/issues/182

Maybe to clarify, this is not a spec or anything special, its just a convention that should bring some DX to graphqlites way of grouping Root Level Mutations with controllers.

andrew-demb:
@Lappihuan 
Here is an example:
```
// directory structure
Controller/
  Session/
    RevokeController.php
    StartController.php
_________________
// RevokeController.php
// ...
use TheCodingMachine\GraphQLite\Annotations\Mutation;

final class RevokeController
{
    #[Mutation(name: 'revokeSession')]
    public function __invoke(): bool
    {
        // just for example
        return true;
    }
}
```

> I'm wondering if you shouldn't just use #[Mutation] as your "namespace" level and use #[Field] for whatever is nested in it.
> Which would already be supported for your case.

Am I be correct if it will require to make some additional layer of classes with logic for "namespaced mutations"?

I think best variant for me will be a possibility to specify `namespace` parameter for `#[Mutation]` over controller method. It should be simple to use and namespace declaration will be near to mutation declaration

oojacoboo:
@Lappihuan makes sense and after reviewing our schema, applying namespaces to mutations is really all that's needed.  We already have around 100 mutations and only a portion of our apps are using GraphQL.  Luckily we have a decent pattern in place, but it's already becoming more problematic to find mutations.

`#[MutationNamespace]` sounds like an appropriate name to me.

https://graphql-rules.com/rules/mutation-namespaces

These list of rules are great overall.  The one on mutation namespaces does a pretty good job explaining how namespaces are helpful for others interested.

@andrew-demb makes a good point about defining namespaces on `#[Mutation]` directly.  But, not necessarily for the reason he mentioned, as that's a fairly strange implementation.  That said, I could easily see cases where you might have mutation/controller methods scattered across various classes that you want grouped together into a single namespace.  I could also see a desire to include a mutation into multiple namespaces.

What if we keep `#[MutationNamespace]` on a class which puts all mutations in that class within the defined namespace.  Then, also allow the attribute to be applied to mutation methods (must have `Mutation` attribute applied).  If the `MutationNamespace` is provided for a mutation method, it's placed within the defined namespaces and not the class' namespace.  To allow for multiple namespaces, the `MutationNamespace` would instead take an array for the property `names`.  We could also keep `name` or even make `name` a `string|array` type for API simplicity.

```php
#[GQL\MutationNamespace(name: 'User')]
class UserController {
    #[GQL\Query]
    public function user()...

    #[GQL\Mutation]
    public function create()...

    #[GQL\Mutation]
    public function update()...

    #[GQL\Mutation]
    #[GQL\MutationNamespace(name: ['User', 'Role'])]
    public function delete()...
```

Lappihuan:
@oojacoboo @andrew-demb 
I'm actually thinking about following the pattern @devmaslov started with `#[Input]` where you can define multiple Namespaces on classes and have a `for` property on `#[Muatation]` and `#[Query]` similar to how `#[Field]` works:
```php
#[GQL\MutationNamespace(name: 'UserMuatations')]
#[GQL\MutationNamespace(name: 'UserQueries')]
#[GQL\MutationNamespace(name: 'Misc')]
class UserController {
    #[GQL\Query(for: "UserQueries")]
    public function user()...

    #[GQL\Query(for: "UserQueries")]
    public function userAlt()...

    #[GQL\Mutation(for: "UserMuatations")]
    public function create()...

    #[GQL\Mutation(for: "UserMuatations")]
    public function update()...

    #[GQL\Mutation(for: ["UserMuatations","Misc"])]
    public function delete()...

    #[GQL\Mutation(for: "Misc")]
    public function computeX()...
```
This gives quite a lot of flexibility and should also integrate with your Invokable Controller pattern as far as i can tell @andrew-demb 

I'd actually also like to add the ability to define multiple `#[Type]` Annotations since i recently was suprised that Type doesn't support that and i really enjoy that feature on Input.

oojacoboo:
@Lappihuan The `MutationNamespace` can be defined anywhere though, right?  You could even include them all in a single namespace file if you wanted?

Lappihuan:
@oojacoboo yes i'd hope so, but i'm still a bit unsure about the implementation of the whole thing.

The output Type annotation https://graphqlite.thecodingmachine.io/docs/annotations-reference#type-annotation

oojacoboo:
Misread you on the `#[Type]` annotations - sounds good there.

As for the API design, I love it 🎉

Lappihuan:
Now i'd actually call it `RequestGroup` or something alike since it can be used for both Mutations and Queries.

oojacoboo:
Mutations and queries are called "operations".  So, `OperationNamespace` would be appropriate.

oojacoboo:
@Lappihuan now that 6.0 is released, it'd be great to get this merged in for 6.1.

oojacoboo:
@Lappihuan hope we haven't lost you.  What do we need to do to get this one merged in?

Lappihuan:
I've since learned this should not be done with the current gql spec: https://github.com/MichalLytek/type-graphql/issues/1219#issuecomment-1055293025

Issue #481, opened 2022-05-24
Parameter is missing for class constructor for boolean properties


ozahorulia:
**Input class:**

```
#[Input]
class SomeInput
{
    public function __construct(
        #[Field(inputType: 'Boolean!')]
        public readonly bool $x,

        #[Field(inputType: 'Boolean!')]
        public readonly bool $y,
    ) {
    }
}
```

**Mutation:**
```
mutation {
  updateSomething(someInput:
    {
      x: false
      y: false
    }
  )
}
```

**Result:**
```
"Parameter 'x' is missing for class 'SomeInput' constructor. It should be mapped as required field."
```

If I change type to any other (e.g. string) - it works. If I pass true instead of false as **x** or **y** argument - it also works.
Looks like problem occurs for boolean fields with false values.

oojacoboo:
@ozahorulia yea, this is very likely.  Are you using `master`?  Also, what's your `Type` annotation look like?  There have been some recent improvements for Input types with a new "update" option.

ozahorulia:
The project is in production, so not master, bust last available version (5.0.2).

My Type class looks like this:

```
#[Type(class: SomeClass::class)]
#[SourceField(name: 'x')]
#[SourceField(name: 'y')]
class SomeClass
{
}
```

And the referenced `SomeClass` itself of course has `getX(): bool` and `getY(): bool` methods.

oojacoboo:
That's understandable @ozahorulia.  We should have a new update released soon, 6.0.  You'll likely need to wait until that happens before this issue is resolved.  However, in the meantime, you might test `master` to see if it resolves your issue, so that we can proceed towards a PR, otherwise.

As you can see in the [docs here](https://graphqlite.thecodingmachine.io/docs/next/annotations-reference#input) there is a new `update` option for `Input` type annotations.  This denotes the `Input` as an update action, and alters the behavior of nullable fields.  I suspect that `false` would be evaluated properly under an "update" `Input` type.

ozahorulia:
As for Type annotation - I have no problems with that, it works just as expected. It's separated from the referenced class just in order to separate models, so it doesn't change anything in case of this issue.

I just tried `#[Input(update: true)]` option, as it exists in `5.0.2`. Unfortunately it does not change anything for me. But it's very clear that the problem appears only if properties are promoted in the constructor.

If I declare it like this, everything works perfectly:

```
#[Input]
class SomeInput
{
[...]

PR #482, opened 2022-05-25
Tests for #410

### LINKED ISSUE #410: Parameter 'locked' is missing for class 'FooInputType' constructor. It should be mapped as required field.
I have a class:

```php<?php

declare(strict_types=1);

namespace App\Graphql\Type\Input;

use App\Domain\Entity\Content\ContentDocument;
use App\Domain\Entity\Content\Group;
use Symfony\Component\Validator\Constraints as Assert;
use TheCodingMachine\GraphQLite\Annotations\Field;
use TheCodingMachine\GraphQLite\Annotations\Input;
use TheCodingMachine\GraphQLite\Types\ID;

#[Input(name: 'FooInputType', default: true)]
class FooInputType
{
    public function __construct(
        #[Field] private string $title,
        #[Field] private bool $locked,
    ) {
    }
}
```

When i send mutation using this input field and set `locked: false` i get an error: 
`Parameter 'locked' is missing for class 'FooInputType' constructor. It should be mapped as required field.`

ozahorulia:
Tests for #411 PR (fixes #410 and #481).

### DIFF
--- a/tests/Fixtures/Inputs/TestOnlyConstruct.php
+++ b/tests/Fixtures/Inputs/TestOnlyConstruct.php
@@ -24,10 +24,17 @@ class TestOnlyConstruct
      */
     private $bar;
 
-    public function __construct(string $foo, int $bar = 100)
+    /**
+     * @Field()
+     * @var bool
+     */
+    private $baz;
+
+    public function __construct(string $foo, bool  $baz, int $bar = 100)
     {
         $this->foo = $foo;
         $this->bar = $bar;
+        $this->baz = $baz;
     }
 
     public function setFoo(string $foo): void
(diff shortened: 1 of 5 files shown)


ozahorulia:
@oojacoboo can you please help to resolve conflicts? I based my fork on the @marhub's outdated branch and I'm not too familiar with the project, so I'm afraid to ruin something :)

ozahorulia:
Hahaha. Okay, so after rabasing I can see that problem seems to be fixed in master already. So this PR probably only adds tests. But it would still be perfect if we could merge it and tag it, so I can require it in my project with a stable tag, rather than on master. Thanks.

PR #500, opened 2022-07-13
Fixed null for union types

### LINKED ISSUE #491: Error with null in union type
I have the following return type `SomeLocation | AnotherLocation | null`
And it throws error 
`TheCodingMachine\GraphQLite\Mappers\CannotMapTypeException: For return type of OrderDeliveryType::location, don't know how to handle type null`

Update: It works with phpdoc. But not with built in php types.

bladl:
Fixes #491


oojacoboo:
@bladl thanks for the PR!

Issue #417, opened 2021-12-06
Targeting PHP ^8.0


oojacoboo:
This post is for planning purposes.  Sometime in the next year or so, PHP 7 will be unsupported, no longer receiving security updates.  It's reasonable that we drop support for PHP 7, only targeting PHP ^8.0, sometime leading up to that date.  PHP 7.4 is the only version that's still receiving security updates, at the moment.  Users that wish to continue using PHP 7.4, can always target a previous version.

The main motivation and reasoning for this, would be to remove support for annotations, entirely.  However, there may be some other refactoring and cleanup that can be done.

- Remove all annotation processing and handling logic
- Remove annotation cache support, as it's no longer needed
- Consider removal of support for `MyCLabs\Enum` in favor of native Enums.

bladl:
Maybe it's time to support php ^8.1 only?
Symfony 6.1 is out, and it has no support for php 8.0 .

oojacoboo:
@bladl why not continue to support PHP 8.0?  We can still support Symfony 6.1 components with PHP 8.0 support as well.  It just means if your target PHP version is PHP 8.0, you won't get Symfony 6.1, but a lower version that we also support.  I'm not aware of any BC issues here.

I do think we could move forward with updates to target a minimum of PHP 8.0 now though.  Security support for PHP 7.4 ends on 28 Nov 2022.

bladl:
Features available in php 8.1 could greatly increase code readability and reduce potential bugs. [ Enums, readonly properties, `array_is_list` and pure intersection types](https://stitcher.io/blog/new-in-php-81#pure-intersection-types-rfc). According to [usage statistics](https://stitcher.io/blog/php-version-stats-july-2022) php 8.1 has greater usage than 8.0 for now and in my opinion with time this break even going to grow faster.

oojacoboo:
@bladl Enums are already implemented and don't require that we drop 8.0 support.  Readonly properties and intersection types are cool, but I don't see any need to implement them requiring an 8.0 drop yet.  Being that this is a widely distributed OSS lib, we have to be more accommodating of user's upgrade restrictions.

I support following PHP's own version support dates:
https://www.php.net/supported-versions.php

Security fix support for 7.4 will end Nov 28th.  At that point, support should be dropped for 7.4.  We should follow the same for all versions, dropping support when security fixes are no longer provided, unless there is a legitimate need to deviate.

oojacoboo:
[...]

PR #508, opened 2022-08-30
Some code improvements

bladl:
#417

### DIFF
--- a/src/Mappers/Root/RootTypeMapperInterface.php
+++ b/src/Mappers/Root/RootTypeMapperInterface.php
@@ -27,19 +27,17 @@ interface RootTypeMapperInterface
 {
     /**
      * @param (OutputType&GraphQLType)|null $subType
-     * @param ReflectionMethod|ReflectionProperty $reflector
      *
      * @return OutputType&GraphQLType
      */
-    public function toGraphQLOutputType(Type $type, ?OutputType $subType, $reflector, DocBlock $docBlockObj): OutputType;
+    public function toGraphQLOutputType(Type $type, ?OutputType $subType, ReflectionMethod|ReflectionProperty $reflector, DocBlock $docBlockObj): OutputType;
 
     /**
      * @param (InputType&GraphQLType)|null $subType
-     * @param ReflectionMethod|ReflectionProperty $reflector
      *
      * @return InputType&GraphQLType
      */
-    public function toGraphQLInputType(Type $type, ?InputType $subType, string $argumentName, $reflector, DocBlock $docBlockObj): InputType;
+    public function toGraphQLInputType(Type $type, ?InputType $subType, string $argumentName, ReflectionMethod|ReflectionProperty $reflector, DocBlock $docBlockObj): InputType;
 
     /**
      * Returns a GraphQL type by name.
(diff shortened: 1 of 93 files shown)


oojacoboo:
Let's remove 7.4 from the CI tests so we can get tests passing.  Also, let's update target PHP version in composer.json and anywhere else I'm not thinking about.  Maybe docs need updating.

bladl:
Composer already has only php >=8.0 dependency. 
Update docs to give only php >=8.0 examples.

bladl:
Can this PR be accepted as starting point for version 6.1?

oojacoboo:
I’m headed to the airport, but I’ll try to find some time to review soon.  We could merge this and other PRs for 6.1 - yes.

oojacoboo:
Just had some time here waiting on my flight; really great work on this - looks awesome.

src/Mappers/Root/RootTypeMapperInterface.php
```diff
@@ -27,19 +27,17 @@ interface RootTypeMapperInterface
 {
     /**
      * @param (OutputType&GraphQLType)|null $subType
-     * @param ReflectionMethod|ReflectionProperty $reflector
      *
      * @return OutputType&GraphQLType
```
oojacoboo, line 31: @bladl was reviewing this further and prepping a release update and noticed inconsistencies on the return typing for this interface.  Was this an oversight, or can you elaborate further here?  Also, the PHPDoc for `$subType` isn't consistent and should probably just be removed right?  Also the return typehint doc?

bladl: I researched existing codebase before making changes. My opinion was that intersection types `OutputType&GraphQLType` and `InputType&GraphQLType` is important here and we should leave PHPDoc as is. 
As I think its not inconsistencies, its clarification. 

We could use [pure intersection types](https://www.php.net/releases/8.1/en.php#pure_intersection_types)  , but they are supported only by PHP 8.1

This PHPDoc with intersection type was added in 2019, and I think we should keep it as is until minimum supported PHP version will be 8.1

oojacoboo: @bladl I see what you're saying, I guess I should clarify some of my questions:

For `toGraphQLInputType` the return type is a union of `InputType|GraphQLType`.  However, `toGraphQLOutputType` the return type is only `OutputType`.  Does the `toGraphQLInputType` support a return type that's a `GraphQLType` and not an `InputType`?

As for the `$subType` parameter, it's similar situation with `toGraphQLInputType` supporting a union and `toGraphQLOutputType only supporting `?OutputType`.

I totally see what you mean about the intersection type and agree with having that PHPDoc.  That clarifies that bit.  Does `toGraphQLInputType` just need to respect the same as `toGraphQLOutputType`?  It seems the typing for `toGraphQLOutputType` is much better.  Where `toGraphQLInputType` seems inaccurate or unnecessarily loose.

bladl: Now I see. You are right!
I will open a PR with fixes soon.

Seems like usage of union type was some kind of workaround for intersection type.

bladl: @oojacoboo Please, review the changes #520

oojacoboo: @bladl awesome thanks!

PR #509, opened 2022-09-01
Code improvements

bladl:
#417

Remove `webmozart/assert` and `phpstan/phpstan-webmozart-assert` library. Replace it with native assertion.

Type safety improvement.


bladl:
I'm sorry, can someone explain why those conflicts appeared?

Issue #517, opened 2022-10-19
Allow psr/container v2


bladl:
GraphqLite is failing to install with my Symfony 6 project. Some of packages require v2 of [psr/container](https://github.com/php-fig/container)
[psr/container](https://github.com/php-fig/container) v2 is required.
Tried to resolve, but [picontainer](https://github.com/thecodingmachine/picotainer) locked psr/container v1.

oojacoboo:
I haven't looked into v2 for `psr/container`.  Is there anything substantial that's changed?

bladl:
> I haven't looked into v2 for `psr/container`. Is there anything substantial that's changed?

Seems like just a boolean type hint was added https://github.com/php-fig/container/releases/tag/2.0.0

bladl:
According to this [deprecation warning](https://github.com/container-interop/container-interop#deprecation-warning). 
[Container intertop](https://github.com/container-interop/container-interop#deprecation-warning) used by [picontainer](https://github.com/thecodingmachine/picotainer)  is already deprecated in favor of psr/container

My proposal is to replace picontainer with a custom one based on psr/container.
Would be glad to make a PR!

bladl:
Seems like this repository is abandoned. Issues not reviewed for a 6 months https://github.com/thecodingmachine/picotainer/issues/8

PR #518, opened 2022-10-21
Psr/container ^2 support

bladl:
Fix https://github.com/thecodingmachine/graphqlite/issues/517


bladl:
Seem like issue with codecov not related to this PR ?

oojacoboo:
@bladl yea, seems like maybe they were having server issues - re-ran tests and everything is passing now.

src/AnnotationReader.php
```diff
@@ -86,7 +87,7 @@ public function __construct(private Reader $reader, private string $mode = self:
      *
      * @template T of object
      */
-    private function getClassAnnotation(ReflectionClass $refClass, string $annotationClass): ?object
```
oojacoboo, line 89: What's wrong with using the nullable, `?`

src/AnnotationReader.php
```diff
@@ -123,7 +124,7 @@ private function getClassAnnotation(ReflectionClass $refClass, string $annotatio
      *
      * @throws AnnotationException
      */
-    private function getMethodAnnotation(ReflectionMethod $refMethod, string $annotationClass): ?object
```
oojacoboo, line 126: What's wrong with using the nullable, `?`

src/AnnotationReader.php
```diff
@@ -231,7 +232,7 @@ static function ($attribute) {
      *
      * @template T of object
      */
-    public function getTypeAnnotation(ReflectionClass $refClass): ?TypeInterface
```
oojacoboo, line 234: What's wrong with using the nullable, `?`

src/AnnotationReader.php
oojacoboo: There are a ton of code changes from using the nullable typing to explicit `null` in a union type definition.  I know PHP 8.0 supports union types.  But, what's the value of this?

andrew-demb: `doctrine/coding-standard` forces using union types for this since https://github.com/doctrine/coding-standard/pull/266

bladl: > There are a ton of code changes from using the nullable typing to explicit `null` in a union type definition. I know PHP 8.0 supports union types. But, what's the value of this?

This changes belongs to commit named `run cs-fix script`. Every line of code was automatically refactored by code fixer configured with [doctrine/coding-standard 10.0](https://github.com/doctrine/coding-standard)

oojacoboo: I guess we can roll with it - sticking with Doctrine's opinions on CS.  Personally, I don't see the appeal.  It's uglier, more confusing to figure out what's being returned from a function (like the main type being expected), and encourages the use of additional return types where a union is already being returned (why not tack on another `|array`, already have `null|Object`).

I'd love to hear some justifications for this decision.  I'm sure it's something along the lines of consistency, being explicit and more obvious that null is returned, where `?` can be overlooked.

src/Containers/LazyContainer.php
```diff
@@ -0,0 +1,58 @@
...
+        $this->delegateLookupContainer = $delegateLookupContainer ?: $this;
+    }
+
+    public function get(string $id): mixed
+    {
+        if (isset($this->objects[$id])) {
+            return $this->objects[$id];
+        }
+        if (! isset($this->entries[$id])) {
+            throw  throw NotFoundException::notFoundInContainer($id);
```
oojacoboo, line 41: `throw new` not `throw throw`

oojacoboo (changes requested):

src/Containers/LazyContainer.php
oojacoboo: Also, can we get a test for this `throw` condition?

bladl: Sure!

oojacoboo (approved):

oojacoboo:
Thanks for all the work on this @bladl

PR #526, opened 2022-11-16
Add php 8.2 to CI

fezfez:


oojacoboo:
@fezfez thanks for submitting this PR!  Having a look at the error here, it looks like we need to make some type changes to on the `ID`.  Implicit type casting has been deprecated now.  Some decisions need to be made here:

Can `ID::val()` return just a `string` value? 
What is the point of a `bool` ID value?
Can we just explicitly typecast the other scalars and rely on `__toString` for objects?  This would allow us to return `string` for `val()`.

We probably need some feedback into this from others.  I've done some searching across our implementation of GraphQL and from what I can see, we're using `ID` strictly as a `string` typed value object.

fezfez:
> @fezfez thanks for submitting this PR! Having a look at the error here, it looks like we need to make some type changes to on the `ID`. Implicit type casting has been deprecated now. Some decisions need to be made here:
> 
>     * Can `ID::val()` return just a `string` value?
> 
>     * What is the point of a `bool` ID value?
> 
>     * Can we just explicitly typecast the other scalars and rely on `__toString` for objects?  This would allow us to return `string` for `val()`.
> 
> 
> We probably need some feedback into this from others. I've done some searching across our implementation of GraphQL and from what I can see, we're using `ID` strictly as a `string` typed value object.

I think bool is a non sens and should be drop,  but it's a BC.
object should also be change by Stringable,  but it's also BC.

Code will be much cleaner but it's for 7.x

src/Types/ID.php
```diff
@@ -35,6 +36,14 @@ public function val(): bool|float|int|string|object
...
-        return (string) $this->value;
+        if (is_bool($this->value)) {
+            return $this->value === true ? '1' : '0';
+        }
+
+        if (is_scalar($this->value)) {
+            return (string) $this->value;
+        }
+
+        return (string) $this->value->__toString();
```
oojacoboo, line 47: `__toString()` should always return a string.  We should remove the explicit type cast.

oojacoboo (changes requested):

src/Types/ID.php
fezfez: done

oojacoboo (approved):

PR #530, opened 2022-11-25
Determine union type name via naming strategy

mdoelker:
Hey there,

this PR moves the hardcoded naming logic of union types constructed from compound PHP types from the UnionType class to NamingStrategy.

I have chosen to pass in the type names that the union consist of as input. If you feel it makes more sense to inject the underlying type objects, let me know.


oojacoboo:
@mdoelker thanks for the PR!  I think it makes sense to stick with the array of names, as you have it.  It follows the pattern used already in the `NamingStrategy`.  This all looks good to me.

Issue #531, opened 2022-11-25
Input: Cached type in registry is not the type returned by type mapper


NormySan:
Hey,

I'm having a weird issue with input types where I get the exception in the title "Cached type in the registry is not the type returned by type mapper".

For some reason, the type stored in the cache is not the same as the one from the type mapper even if the input class they both reference is the same.

The mutation that I try to run only fails when I use variables to run the mutation, If I run the mutation without variables and provide the values directly it all works but this is not possible to do in the frontend since everything is automatically generated and requires the type to be provided.

**Failing mutation**

```graphql
mutation CreateFacility($input: FacilityInput!) {
  createFacility(input: $input) {
    id
  }
}
```

**Working mutation**

```graphql
mutation CreateFacility {
  createFacility(input: {
    name: "Facility"
    subdomain: "facility"
  }) {
    id
  }
}
```

**Input type**

```php
#[Input]
class FacilityInput
{
    #[Field]
    public string $name;

    #[Field]
    public string $subdomain;
}

```

**Controller/Resolver**

```php
class FacilityResolver
{
    public function __construct(
        private readonly CommandBus $commandBus,
    ) {
    }

    #[Mutation]
    #[Logged]
    public function createFacility(
        #[InjectUser] Employee $employee,
        FacilityInput $input
    ): Facility {
        /** @var Facility $facility */
        $facility = $this->commandBus->dispatch(new CreateFacilityCommand(
            (string) $employee->organizationId,
            $input->name,
            $input->subdomain,
        ));

        return $facility;
    }
}
```

**The cached type** 

```
object(TheCodingMachine\GraphQLite\Types\TypeAnnotatedObjectType)[782]
  public 'name' => string 'FacilityInput' (length=13)
  public 'description' => null
  public 'astNode' => null
  public 'config' => 
    array (size=3)
      'name' => string 'FacilityInput' (length=13)
      'fields' => 
        object(Closure)[775]
          virtual 'closure' 'TheCodingMachine\GraphQLite\Types\TypeAnnotatedObjectType::TheCodingMachine\GraphQLite\Types\{closure}'
          public 'static' => 
            array (size=5)
              ...
      'interfaces' => 
        object(Closure)[842]
          virtual 'closure' 'TheCodingMachine\GraphQLite\Types\TypeAnnotatedObjectType::TheCodingMachine\GraphQLite\Types\{closure}'
          public 'static' => 
            array (size=3)
              ...
  public 'extensionASTNodes' => 
[...]

PR #532, opened 2022-12-15
Fix AnnotationReader returning Input classes as Type classes.

mdoelker:
This fixes the issue described in #531 that would result in error "Cached type in registry is not the type returned by type mapper" when using Input types in mutations via variables.

### DIFF
--- a/src/AnnotationReader.php
+++ b/src/AnnotationReader.php
@@ -235,8 +235,7 @@ class AnnotationReader
     public function getTypeAnnotation(ReflectionClass $refClass): TypeInterface|null
     {
         try {
-            $type = $this->getClassAnnotation($refClass, Type::class)
-                ?? $this->getClassAnnotation($refClass, Input::class);
+            $type = $this->getClassAnnotation($refClass, Type::class);
 
             if ($type !== null && $type->isSelfType()) {
                 $type->setClass($refClass->getName());


oojacoboo:
Thanks for the PR @mdoelker.  

So, I think this change is needed, at least according to how it should probably be written.  However, I tested your branch on our repo and it's actually failing.  

It seems this method is being called on some `Input` annotated (attribute in our case) types: https://github.com/thecodingmachine/graphqlite/blob/master/src/TypeGenerator.php#L47

Here is the call stack:
```
TheCodingMachine\GraphQLite\MissingAnnotationException: GraphQL type class "Foo\Input" must provide a @Type annotation.

/srv/www/vendor/thecodingmachine/graphqlite/src/MissingAnnotationException.php:18
/srv/www/vendor/thecodingmachine/graphqlite/src/TypeGenerator.php:50
/srv/www/vendor/thecodingmachine/graphqlite/src/Mappers/AbstractTypeMapper.php:264
/srv/www/vendor/thecodingmachine/graphqlite/src/Mappers/CompositeTypeMapper.php:65
/srv/www/vendor/thecodingmachine/graphqlite/src/Mappers/RecursiveTypeMapper.php:300
/srv/www/vendor/thecodingmachine/graphqlite/src/Mappers/RecursiveTypeMapper.php:359
/srv/www/vendor/thecodingmachine/graphqlite/src/Mappers/Root/BaseTypeMapper.php:71
/srv/www/vendor/thecodingmachine/graphqlite/src/Mappers/Root/EnumTypeMapper.php:60
/srv/www/src/Acme/GraphQL/Type/Mapper.php:128
/srv/www/vendor/thecodingmachine/graphqlite/src/Mappers/Root/CompoundTypeMapper.php:54
/srv/www/vendor/thecodingmachine/graphqlite/src/Mappers/Root/IteratorTypeMapper.php:45
/srv/www/vendor/thecodingmachine/graphqlite/src/Mappers/Root/NullableTypeMapperAdapter.php:55
/srv/www/vendor/thecodingmachine/graphqlite/src/Mappers/Parameters/TypeHandler.php:350
/srv/www/vendor/thecodingmachine/graphqlite/src/Mappers/Parameters/TypeHandler.php:88
/srv/www/vendor/thecodingmachine/graphqlite/src/FieldsBuilder.php:401
/srv/www/vendor/thecodingmachine/graphqlite/src/FieldsBuilder.php:302
/srv/www/vendor/thecodingmachine/graphqlite/src/FieldsBuilder.php:208
/srv/www/vendor/thecodingmachine/graphqlite/src/Types/TypeAnnotatedObjectType.php:44
/srv/www/vendor/webonyx/graphql-php/src/Type/Definition/FieldDefinition.php:97
/srv/www/vendor/webonyx/graphql-php/src/Type/Definition/TypeWithFields.php:26
/srv/www/vendor/webonyx/graphql-php/src/Type/Definition/TypeWithFields.php:77
/srv/www/vendor/thecodingmachine/graphqlite/src/Types/MutableTrait.php:124
/srv/www/vendor/thecodingmachine/graphqlite/src/Types/MutableTrait.php:83
/srv/www/vendor/webonyx/graphql-php/src/Utils/TypeInfo.php:202
/srv/www/vendor/webonyx/graphql-php/src/Utils/TypeInfo.php:222
/srv/www/vendor/webonyx/graphql-php/src/Type/Schema.php:234
/srv/www/vendor/webonyx/graphql-php/src/Type/Schema.php:208
/srv/www/vendor/webonyx/graphql-php/src/Type/Introspection.php:240
/srv/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php:623
/srv/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php:550
/srv/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php:1195
/srv/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php:1145
/srv/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php:1106
/srv/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php:793
/srv/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php:741
/srv/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php:654
/srv/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php:557
/srv/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php:1195
/srv/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php:264
/srv/www/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php:215
/srv/www/vendor/webonyx/graphql-php/src/Executor/Executor.php:156
/srv/www/vendor/webonyx/graphql-php/src/GraphQL.php:169
/srv/www/vendor/webonyx/graphql-php/src/Server/Helper.php:311
/srv/www/vendor/webonyx/graphql-php/src/Server/Helper.php:211
/srv/www/vendor/webonyx/graphql-php/src/Server/StandardServer.php:136
/srv/www/vendor/webonyx/graphql-php/src/Server/StandardServer.php:171
```

I'm not sure why it's calling the `TypeGenerator` here instead of the `InputTypeGenerator`, or even if the `InputTypeGenerator` should be called.  I'd expect the `RecursiveTypeMapper` to be detecting the `Input`, but for some reason it's not.

In any case, it seems this is the reason this was put in place, which clearly looks like WIP code.  I'll try to find some time to dig deeper into it.

oojacoboo:
So, I've been digging into this and the issue is that `TypeHandler::mapReturnType` receives both input and output types.  However, there isn't currently a way to determine if it's an input or output and that's needed to map it accordingly.

Creating a testcase for this is difficult because it seems very dependent on the order in which the schema builder processes types.  I believe it's using lexical ordering from the registered type namespaces.  However, I'm not 100% on this.  I am aware of another issue where there are ordering issues with `Input` type defined classes and inputs built with a factory.

Working on a fix for this and hopefully I can get something reasonable put together.  I don't really like how convoluted it is right now to determine what types you're working with.  In reality, these type mappers shouldn't be dealing with these reflection classes and should, instead, be using a proxy class we manage with methods that return reliable details about the the type (input, output, nullable, etc.).

bladl:
>  I am aware of another issue where there are ordering issues with `Input` type defined classes and inputs built with a factory.
> 

Can the same reason cause issue https://github.com/thecodingmachine/graphqlite/issues/536?

oojacoboo:
@bladl could absolutely be related.  As it's currently written, the order in which types are processed matters.  It obviously does a pretty good job, but there are clearly some holes.  Making sense of all these type mappers, IMO, is complicated and frustrating.  And I blame that on the interface design, coupled with a lack of documentation and sub-optimal parameter names.

oojacoboo:
@mdoelker Yep, that's basically it.  So, it appears the issue is with the factories requiring a `Type` annotation.  When a field, of what's actually an input type, returns another input type - there in lies the issue.  Using factories and Input annotated classes together just doesn't work well.  I don't know why factories are requiring a `Type` annotation, except for legacy reasons, since they were the initial way of creating input types before the `Input` annotation was added.

I guess it's possible that could be changed to require an `Input` annotation instead and maybe they'll work well together.

oojacoboo:
I've created #537 for discussion on factory updates.  I see no reason this PR shouldn't be merged, even if it's a breaking change for some.

Issue #540, opened 2023-01-09
Webonyx v15


bladl:
`webonyx` v15.0 is out. 
https://github.com/webonyx/graphql-php/releases/tag/v15.0.0

Will be both v14 and v15 supported by `graphqlite`? At first glance there is backward-incompatible changes

Bumping a new major version of `webonyx` requires updating [ecodev/graphql-upload](https://github.com/thecodingmachine/graphqlite-bundle) first.

oojacoboo:
[ecodev/graphql-upload](https://packagist.org/packages/ecodev/graphql-upload) doesn't support webonyx v15 yet.  So, we would need to wait for support there first.

Outside of that, their update notes don't make it very clear what's likely to be broken... I guess anything "changed".  But there is a lot of smaller things there that aren't likely to be a BC break.

We could start with a PR that updates the dep, see what's breaking and go from there.

PowerKiKi:
`ecodev/graphql-upload` has been released with `webonyx/graphql-php` 15 support: https://github.com/Ecodev/graphql-upload/releases/tag/7.0.0

oojacoboo:
@PowerKiKi thanks for that!
@bladl want to give a PR a shot, updating the deps - see what we're looking at?

oojacoboo:
@PowerKiKi was there a specific reason for PHP 8.1 minimum support and not 8.0?

PowerKiKi:
All my projects are on PHP 8.1, and I am not not keen of the extra mental burden to support multiple PHP versions. I don't even have PHP 8.0 installed anymore. And 8.0 will be EOL quite soon in 10 months, so I don't think it's doing anyone any favor to start using something that will soon be deprecated.

According to https://packagist.org/php-statistics, PHP 8.0 is "only" ~25% of all PHP 8+, and I expect it would go decrease even more with the adoption of the recent 8.2 in the coming months.

oojacoboo:
@bladl @PowerKiKi thanks guys!  It's been merged.

PR #542, opened 2023-01-12
graphql-php v15

bladl:
We have 23 test errors after bumping webonyx version
https://github.com/thecodingmachine/graphqlite/issues/540


bladl:
@oojacoboo Minimum supported php version for ecodev/graphql-upload is ^8.1. But we have continuous integration for php 8.0

bladl:
Other test errors I will try to fix by myself

bladl:
There was a bunch of breaking changes. 
Finally, all php >8.0 tests passed. Please review a PR !

oojacoboo:
Dropped 8.0 tests.  I'll review the rest of the PR soon.

bladl:
Does it means we drop support for php 8.0? So we can use all range of php 8.1 features from now

oojacoboo:
Yes, we'd basically be dropping 8.0 support and only supporting >=8.1 with this PR.  However, I am wondering why `ecodev/graphql-upload` has minimum 8.1 support.

Also, I didn't check.  What were the test errors surrounding 8.0?

bladl:
> 
> 
> Also, I didn't check.  What were the test errors surrounding 8.0?

composer install failed so I did not run tests in php 8.0 enviroment.

bladl:
While I was working for this and previous PR I felt a huge lack of native PHP [intersection types](https://php.watch/versions/8.1/intersection-types). Many docblocks and assertions can be dropped with php 8.1. 

With all range of php 8.1 features debugging codebase would be much simpler and I believe graphqlite will evolve much faster.
In my opinion it's worth to drop support for php 8.0 now in any case.

PR #543, opened 2023-01-12
Preserve type already added

fezfez:
Hello,

With this PR this code will preserve MyCustomInputType

```php
        $factory = new SchemaFactory();
        $factory->setSchemaConfig(SchemaConfig::create([
            'types' => [
                MyCustomInputType::getInstance()
            ]
        ]));

       $schema = $factory->createSchema();
```

### DIFF
--- a/src/Schema.php
+++ b/src/Schema.php
@@ -79,7 +79,7 @@ class Schema extends \GraphQL\Type\Schema
         if (is_callable($preservedType)) {
             $preservedType = $preservedType();
         }
-        $config->setTypes([...$preservedType, ...$recursiveTypeMapper->getOutputTypes()]);
+        $config->setTypes(array_merge($preservedType, $recursiveTypeMapper->getOutputTypes()));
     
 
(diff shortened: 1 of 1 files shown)


fezfez:
Is this a good approach ? if it seem to good to you i will fix the tests, otherwise can you provide a good way to do this ?

oojacoboo:
@fezfez Can you describe in a bit more detail what it is you're hoping to accomplish here?  What's the design challenge you're having in your API?  Is it that you're basically wanting a static type?  Assuming so, I don't see any point in doing this, as opposed to just programming an input to be static itself.

fezfez:
@oojacoboo i have created a very lite sample of my problem at https://github.com/fezfez/sample-graphqlite

oojacoboo:
@fezfez , would it be possible to get an explanation of your intended goals here as well?  I can see from the code you're trying to create custom types using the webonyx API.  But why?  Can you explain why you cannot use a normal annotated object, or what the overall goal is?

oojacoboo:
Closing unless this is still needed.

PR #561, opened 2023-02-28
fix Enum not resolved in first level input arguments

### LINKED ISSUE #536: Enum not resolved in first level input arguments
Consider such queries that do the same thing - update order status.

Schema
```
enum OrderState {
	Completed
	Failed
	Processing
}
input OrderInput {
	id: String!
}
input OrderManagementUpdateInput {
	state: OrderState = null
}
type Schema {
  updateShopOrderPrimaryData(
    order: OrderInput!
    update: OrderManagementUpdateInput!
  ): Order!
}
```
Query 1
```
mutation updateShopOrderState(
    $orderId: String!
    $state: OrderState!
) {
    updateShopOrderPrimaryData(order: {id: $orderId}, update: {state: $state}) {
        ...DashboardOrderFields
    }
}
```

Query 2
```
mutation updateShopOrderPrimaryData(
    $orderId: String!
    $update: OrderManagementUpdateInput!
) {
    updateShopOrderPrimaryData(order: {id: $orderId}, update: $update) {
        ...DashboardOrderFields
    }
}
```

Actually, issue is that Query 1 throws exception `cannot find GraphQL type \"OrderState\". Check your TypeMapper configuration.`
Query 2 works just fine when I pass `state` field in `update` argument .

Actually its the same queries, but they have different client-side structure.

I tried to make reproduction with tests https://github.com/thecodingmachine/graphqlite/pull/535 but it seems issue not so obvious.

Lappihuan:
Fixes  #536

The mentioned merge in the ticket above doesn't actually fix this.

Due to a missing ltrim not being carried over from the MyClabsEnumTypeMapper the FQCN of the enum most of the time had a leading backslash resulting it not being found in the cache or duplication once with and once without leading backslash.

Excluding of enums for Types is not done on the NS getClassList method but rather later in the AbstractTypeMapper so EnumTypeMapper can actually find Enums (this whole implementation was previously working by accident)

The existing TestCase only worked because the same Enum was used as Input directly without passing them as variables, removing that testcase resulted in the singleEnum mutation to fail with the typical "check your typemapper configuration".

We also added a additional TestCase to use variables for Types since variables are probably what a client like ApolloClient would send, but allmost all testcases just have the raw values directly in the document which bypasses a lot of graphql-php validation rules.

### DIFF
--- a/src/Mappers/AbstractTypeMapper.php
+++ b/src/Mappers/AbstractTypeMapper.php
@@ -30,6 +30,7 @@ use TheCodingMachine\GraphQLite\Types\MutableInterfaceType;
 use TheCodingMachine\GraphQLite\Types\MutableObjectType;
 use TheCodingMachine\GraphQLite\Types\ResolvableMutableInputInterface;
 
+use UnitEnum;
 use function assert;
 
 /**
@@ -114,6 +115,13 @@ abstract class AbstractTypeMapper implements TypeMapperInterface
         $classes = $this->getClassList();
 
         foreach ($classes as $className => $refClass) {
+            // Enum's are not types
+            if (interface_exists(UnitEnum::class)) {
+                // @phpstan-ignore-next-line - Remove this after minimum supported PHP version is >= 8.1
+                if ($refClass->isEnum()) {
+                    continue;
+                }
+            }
             $annotationsCache = $this->mapClassToAnnotationsCache->get($refClass, function () use ($refClass, $className) {
                 $annotationsCache = new GlobAnnotationsCache();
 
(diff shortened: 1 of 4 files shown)


src/Mappers/AbstractTypeMapper.php
```diff
@@ -114,6 +115,13 @@ private function buildMap(): GlobTypeMapperCache
         $classes = $this->getClassList();
 
         foreach ($classes as $className => $refClass) {
+            // Enum's are not types
+            if (interface_exists(UnitEnum::class)) {
+                // @phpstan-ignore-next-line - Remove this after minimum supported PHP version is >= 8.1
```
oojacoboo, line 120: `master` is already targeting >= 8.1

src/Mappers/AbstractTypeMapper.php
```diff
@@ -186,6 +194,13 @@ private function buildMapClassToExtendTypeArray(): GlobExtendTypeMapperCache
 
         $classes = $this->getClassList();
         foreach ($classes as $refClass) {
+            // Enum's are not types
+            if (interface_exists(UnitEnum::class)) {
```
oojacoboo, line 198: `master` is already targeting >= 8.1

tests/Integration/EndToEndTest.php
```diff
@@ -1175,6 +1175,63 @@ public function testEndToEndEnums3(): void
         $this->assertSame(['echoProductType' => 'NON_FOOD'], $this->getSuccessResult($result));
     }
 
+    /** @requires PHP >= 8.1 */
```
oojacoboo, line 1178: `master` is already targeting >= 8.1

oojacoboo (changes requested):

oojacoboo:
Thanks for the PR!  We're already targeting >= 8.1 now.  If you could remove those additional conditions that'd be great.

oojacoboo (approved):

PR #564, opened 2023-03-02
Fix MagicField & SourceField to include $annotations parameter

### LINKED ISSUE #563: Missing `annotations` parameter in #[MagicField] and #[SourceField]
Both `#[MagicField]` and `#[SourceField]` accept annotations that should be passed through due to inability to specify the attributes directly on the field:

```php
#[Type]
#[MagicField(attributes: ['annotations' => [new MyCustomMiddleware]], name: 'field')]
class MyType {}
```

All other properties of `MagicField` and `SourceField` can be specified directly, without using the "annotations API" (which is `$attributes`), so I believe them missing a `array $annotations = []` parameter in the end is a bug:

```php
#[Type]
#[MagicField(name: 'field', annotations: [new MyCustomMiddleware])]
class MyType {}
```

Confirm this is not the expected behaviour and I'll make a PR with a fix, thanks!

oprypkhantc:
Closes #563

### DIFF
--- a/src/Annotations/MagicField.php
+++ b/src/Annotations/MagicField.php
@@ -46,8 +46,11 @@ class MagicField implements SourceFieldInterface
     /** @var array<string, ParameterAnnotations> */
     private $parameterAnnotations;
 
-    /** @param mixed[] $attributes */
-    public function __construct(array $attributes = [], string|null $name = null, string|null $outputType = null, string|null $phpType = null, string|null $description = null, string|null $sourceName = null)
+    /**
+     * @param mixed[] $attributes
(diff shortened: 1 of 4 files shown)


oprypkhantc:
PHPStan is failing but it seems like the problem is because `phpstan` is not locked to a patch/minor version in `composer.json`. Same goes `doctrine/coding-standard` locally, by the way - it spits out a couple of errors completely unrelated to my changes.

Maybe just lock those two to a specific minior version? `phpstan: "~1.9.0"` and `doctrine/coding-standard: "~11.0.0"`

oojacoboo:
These errors should be resolved on `master` now.  I pushed commits that fixed them yesterday.  I'm not sure why they're failing again here.

PR #570, opened 2023-03-08
Fix handling of native union types containing null

### LINKED ISSUE #567: Field nullable union type is throwing an error
Having a type like this:

```php
#[Type]
class File {
	#[Field(name: 'sizes')]
	public function sizes(): ImageSizes|VideoSizes|null {}
}
```

throws an error saying `null` can't be converted to a GraphQL type. GraphQL spec does allow nullable unions, so this seems to be a bug with `graphqlite`. 

For the above function `phpDocumentor` reports the following type:
```php
new Nullable(
	realType: new Compound(
		types: [
			new Object_(),
			new Object_(),
			new Null_(),
		]
	)
)
```
so when `NullableTypeMapperAdapter` attempts to get a non-nullable version of that it goes through https://github.com/thecodingmachine/graphqlite/blob/master/src/Mappers/Root/NullableTypeMapperAdapter.php#L129
and returns the `Compound` type instance with `Null_` being one of the types.

Fix should be trivial: return `return $this->getNonNullable($type->getActualType())` instead of `return $type->getActualType()`

I can PR this change.

oprypkhantc:
Closes #567

TypeHandler wraps native types in `Nullable` if they allow null, even if it's a union with `null` being one of the types. I figured a fix in `NullableTypeMapper` would be needed anyway, in case someone wrongfully does the same, and haven't touched the `TypeHandler` to avoid recursively iterating through types in it.

### DIFF
--- a/src/Mappers/Root/NullableTypeMapperAdapter.php
+++ b/src/Mappers/Root/NullableTypeMapperAdapter.php
@@ -126,7 +126,7 @@ class NullableTypeMapperAdapter implements RootTypeMapperInterface
             return null;
         }
         if ($type instanceof Nullable) {
-            return $type->getActualType();
+            return $this->getNonNullable($type->getActualType());
         }
         if ($type instanceof Compound) {
             $types = array_map([$this, 'getNonNullable'], iterator_to_array($type));
--- a/tests/Mappers/Root/NullableTypeMapperAdapterTest.php
+++ b/tests/Mappers/Root/NullableTypeMapperAdapterTest.php
@@ -2,6 +2,7 @@
 
 namespace TheCodingMachine\GraphQLite\Mappers\Root;
 
+use Generator;
 use GraphQL\Type\Definition\InputType;
 use GraphQL\Type\Definition\NamedType;
 use GraphQL\Type\Definition\NonNull;
@@ -10,6 +11,7 @@ use GraphQL\Type\Definition\StringType;
 use GraphQL\Type\Definition\Type as GraphQLType;
 use phpDocumentor\Reflection\DocBlock;
 use phpDocumentor\Reflection\Type;
+use phpDocumentor\Reflection\Types\Nullable;
 use ReflectionMethod;
 use TheCodingMachine\GraphQLite\AbstractQueryProviderTest;
 use TheCodingMachine\GraphQLite\Fixtures\TestObject;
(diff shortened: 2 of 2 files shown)


oprypkhantc:
By the way, code is not formatted correctly because phpcs still fixes way too much files for this small PR, and there's no GitHub action workflow for it. 

I'll add a github workflow and fix the codestyle in a separate PR, so let's hold this before codestyle is merged

oprypkhantc:
@oojacoboo Nevermind, I can't get `phpcs --tab-width=4` to work, even though it's officially supported and is in the docs, it still doesn't replace tabs with spaces.

What do you think of [php-cs-fixer](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer)? It's very similar to php codesniffer but at least it fixes tabbing. It has a ton of fixers with a lot of configuration and ready rule sets - for PSR2, PSR12, Symfony and Doctrine annotations. It's also the one used by `symfony/symfony`.

If you don't mind, I can PR php-cs-fixer improvement as well. It's mostly plug-n-play.

oojacoboo:
@oprypkhantc thanks for this.  

On `php-cs-fixer`, it can be a great tool.  I haven't dug into the phpcs configuration with this repo much.  Can't you just use `php-cs-fixer` on the `phpcs.xml.dist` config?  I don't see any reason to add the dependency to the repo when the tool can be used as needed by the developer.

src/Mappers/Root/NullableTypeMapperAdapter.php
```diff
@@ -126,7 +126,7 @@ private function getNonNullable(Type $type): Type|null
             return null;
         }
         if ($type instanceof Nullable) {
-            return $type->getActualType();
+            return $this->getNonNullable($type->getActualType());
```
oojacoboo, line 129: Does this read weird, or am I missing something here?

If $type is an instance of nullable, get non-nullable?

oprypkhantc:
@oojacoboo `php-cs-fixer` can't be used with `phpcs`'s config, unfortunately. They have their own format with their own set of fixers & configuration.

Right now the code style is not enforced in any way - it's skipped in the Github workflow, so anything can be mistakenly merged with an invalid codestyle. Having a consistent fixer/checker would be nice for newcomers like me who are used to different codestyle.

I'll PR a change of `phpcs` to `php-cs-fixer` then, if you don't like - feel free to decline it.

oojacoboo (approved):

PR #574, opened 2023-03-10
Add support for void return type

### LINKED ISSUE #573: First party support for `void` return type
Currently if you try to return `void` from a mutation (as it the case with "delete" type of mutations), you get an error for being unable to map the `Void_` type. Because this is a common and valid use case for PHP, I believe it should be supported by `graphqlite` natively, the same way as `DateTime` is natively supported:

```php
#[Mutation]
public function deleteSomeEntity(SomeEntity $entity): void
{
    $entity->delete();
}
```

It's a somewhat [common practice](https://the-guild.dev/graphql/scalars/docs/scalars/void) to add a scalar type `Void` to handle such cases. This can be done in `graphqlite` too with a root type mapper.

As always, if this is desirable, I can PR this improvement. Thoughts?

oprypkhantc:
Closes #573

On the implementation: I've implemented `VoidTypeMapper` separately from the rest to make sure it only ever matches standalone output types, i.e. that it can't be used in a union or as an input or anything else that doesn't make sense. `NullableTypeMapper` is no longer the "top root type mapper" - that's because void type must not be wrapped in `NonNull` as that causes an error when actually running a mutation with `void` return.


src/Mappers/Root/VoidTypeMapper.php
```diff
@@ -0,0 +1,52 @@
+<?php
```
oojacoboo, line 1: So, I'm a little confused on the purpose of the `src/Types.php` file.  I seems like you're creating this as a means of separating out the GraphQLite "custom-ish" types.  But, I don't really see how these differ much from the other types in the `Types` dir.  They're all internally managed types and pulling out a few seems a bit confusing.

Is the internal caching here providing much value?  The instantiation of these objects seems trivial.

src/Mappers/Root/LastTopRootTypeMapper.php
```diff
@@ -0,0 +1,42 @@
...
+use GraphQL\Type\Definition\Type as GraphQLType;
+use phpDocumentor\Reflection\DocBlock;
+use phpDocumentor\Reflection\Type;
+use ReflectionMethod;
+use ReflectionProperty;
+
+/**
+ * The last "top" root type mapper that always calls the dynamicly set "next" mapper.
+ */
+class LastTopRootTypeMapper implements RootTypeMapperInterface
```
oojacoboo, line 19: Does it make sense to call it the last "top" "root" type mapper?  Or does LastTypeMapper or LastRootTypeMapper make more sense?  The name does help with understanding the dynamic nature.  However, it seems like we might be able to add further clarity in the naming?  Is it really the "top" if it's the "last" and dynamic?

oprypkhantc: Agreed. What about "LastMutableTypeMapper" or "LastDelegatingTypeMapper"? On the "Root" suffix, all other root mappers don't have for some reason so I think I should remove it.

src/Mappers/Root/VoidTypeMapper.php
oprypkhantc: The idea is similar to webonyx's `Type` class with `string`, `id`, `getNamedType`, `union` etc method, i.e. if you need a type you know right away where to get it from. The value of `Types` in this case is the same kind of "entry point" for those who need those types in 3rd party type mappers. I'm not worried about caching so much.

For example, in our project we use `carbon/carbon` instead of php's `DateTime`, but we still use `graphqlite`'s built in `DateTime` type to represent Carbon dates it in the Schema and serialize/deserialize them. When more types are added into `graphqlite`, it will always be easy to navigate and find them.

If you're against providing that kind of public API I'll remove it.

src/Mappers/Root/LastTopRootTypeMapper.php
oojacoboo: `LastDelegatingTypeMapper` makes more sense to me.

src/Mappers/Root/VoidTypeMapper.php
oojacoboo: I see.  For DateTime, we use a custom type mapper for our internal `Date` object that maps to a custom `Date` scalar.  We do allow GraphQLite to handle `\DateTime` and `\DateTimeImmutable`.  We have a few other custom types as well that we manage through our type mapper.

Carbon extends `\DateTimeImmutable` IIRC.  I think GraphQLite should handle the serialization of this already.  Of course, you could setup a type mapper to handle the deserialization side as well.

I'm trying to think where this API would even be helpful to a normal developer.  And internally, I don't see much benefit over using the type classes directly.  You can't type with this interface, requiring the class to be typed directly, either.  I think I'd rather we stick with the type classes directly, for the sake of simplicity and consistency.

oprypkhantc: Following the style of `BaseTypeMapper`, I still used a static property to hold an instance of `VoidType`, just to avoid the unnecessary instantiations.

PR #576, opened 2023-03-11
Replace PHP CodeSniffer with php-cs-fixer

oprypkhantc:
I've resembled project's actual code style with php-cs-fixer's rules, aside from a few differences here and there. Everything you see is configurable, so please comment on things you don't like and I'll change them. 

I've also added a CI job to make sure the code style is strictly enforced and will never become out of sync again. It's not under the regular php-versions matrix because code style will always be checked the same way, regardless of php-version or dependencies, so it doesn't make sense to run it 4 times instead of 1.


oprypkhantc:
@oojacoboo On the line length: unfortunately it can't be strictly enforced. However, I did ensure that if you do this:
```php
function something(
    $something1, $something2, $something3,) // etc out of line length
```
or this
```php
something(
    1, 2, 3, 4, 5
)
```
it will be enforced as multiline.

oojacoboo:
Okay, I've had a read over this and done some additional research, etc.  I've used `php-cs-fixer` in the past, although I've mostly used `phpcs`.  It seems the main difference here is that `php-cs-fixer` is designed to actually "fix" the code, whereas `phpcs` is mostly designed to statically analyze the code.  Obviously with `php-cs-fixer`, you have the whole `dry-run` approach.  I wasn't thinking about that being used in a CI pipeline, but makes sense.

It's worth pointing out that `phpcs` has `phpcsbf` which will actually "fix" the code, like `php-cs-fixer` based on the `phpcs` config.

The real question is... which tool is better.  I'm having a hard time understanding how `php-cs-fixer` makes sense in a CI pipeline, when the primary design principle is around "fixing" the code.  In a normal operation, you don't want to automagically fix your code, you want to analyze and report the issues.  It's maybe nice to click a button and have it do that - although I'm not a huge fan of this method, as a developer on a normal basis.  It's great to get something inline initially.

The line-length issue is a prime example of where `php-cs-fixer`'s design thesis fails.  They refuse to support this due to the difficulty of implementation.  That's because the tool is designed to fix this and knowing where to automagically "fix" the code for line length violations, is obviously challenging.

Thoughts?  Why not stick with `phpcs`, maybe clean up some of the rules a bit and run `phpcsbf`?  What advantages would `php-cs-fixer` provide?

oprypkhantc:
Now that you've said it, it all makes sense: 

The difference here is `code sniffer` attempts to not only be a code style fixing tool, but also a code _quality_ reporting tool, which are two entirely different things. `phpcs` has sniffers like `SuperfluousInterfaceNaming` which are not about the code style at all and `phpcbf` will never be able to fix that automatically as it requires scanning and changing the whole codebase. `php-cs-fixer` doesn't have one because it can't be fixed automatically.

---

On the `php-cs-fixer`:

So we're a team of 9 devs, all of whom use `php-cs-fixer` daily. It's in our pre-commit - you can't even commit code without it being correctly formatted. We've been using it for more than 3 years already and while there were a few nuances here and there, it's been incredibly reliable in a sense that we _do_ want it to automatically magically fix our code. Long gone are the days where codestyle issues were pointed out on pull requests and fixed manually.

So that's one thing where `php-cs-fixer` shines - fixing everything automatically, because 99% of things can be fixed automatically. With `phpcs` there are still code style things that you have to fix by hand, which really goes against our principles not to do anything by hand if it can be automated. I assume this project is mostly driven by the same principle - it has CI with checks, it guesses lots of things from the codebase. 

You are correct about the line length, but in a context of code _style_, this seems to be the only thing that is not implemented in `php-cs-fixer` compared to `phpcs`. Otherwise `php-cs-fixer` has a lot more fixers and configuration which again is useful when enforcing the code style.

---

My main motivation for using `php-cs-fixer` was that in terms of code style, it's a better tool. I've never thought of code quality being an issue but it might be. In that context, I've researched a bit and found [phpinsights](https://phpinsights.com/insights/code.html). It's just a fancy wrapper tool that combines `php-cs-fixer` and `phpcs` together. And the reason they do it is simple: they target both code style and code quality at the same time.

So instead of replacing `phpcs` and instead of missing out on features of `php-cs-fixer`, maybe use both? One for code quality, the other one for code style. One can only be fixed by hand, the other can be fixed automatically. Though I'm not familiar with `phpinsights`, it's a tool from `nunomaduro` and certainly is popular (3mil downloads), so it might be worth a try. 

What do you think?

oojacoboo:
Thanks for the detailed input on this and your experience.  That does help put some things in perspective.  

I just had a look over `phpinsights` and it looks great.  It even supports a GitHub actions workflow to annotate issues in a PR.  I think it's worth a shot if we can minimize the overlapping of checks and find a way to cleanly separate the CS/CQ checks.

Do you use `php-cs-fixer` within your IDE on key up, after delay or on save?  If so, is it a dry-run that's reporting the CS issues, or is it actually reformatting the code in the editor continually?

oprypkhantc:
It depends. So speaking of PHPStorm it has first-party support for it and allows any configuration:
 - inspection only, where it underlines the code. I myself don't like it because it has a delay, and I don't know of anyone in our team who uses that
 - run php-cs-fixer (fix, not dry-run) on reformat or on custom keybind - that's what most people use, it works okay
 - run reformat (alongside php-cs-fixer) on save - it's only become available recently. Again works okay
 
 I myself use the last option, but it's a bit buggy yet. Works especially bad when refactoring as it triggers `php-cs-fixer` for each file separately :(

oprypkhantc:
I'll reopen once I get the hang of phpinsights.

oprypkhantc:
I'll reopen once I get the hang of phpinsights.

PR #584, opened 2023-03-20
Immutability

### LINKED ISSUE #581: Mutability concerns
Currently many of the classes are mutable, including fields and types. While this has been working so far for us, we've never actually tested it on production so there might be bugs related to it. And unfortunately, even if there are none, graphqlite doesn't (and shouldn't) have tests coverage for it so it might become an issue in the future after any changes.

We've had prior instances where a previous (mutable) state under very specific conditions was shared between multiple requests/jobs and you have to go through heaps of hell to debug it, fix it and hope the business wasn't affected too much. That's why I'm worried about `graphqlite`'s usage of mutability.

This is somewhat fine for `SchemaFactory` and field descriptors, because it's only executed once prior to handling requests, but with everything else - i.e. types & fields, this is a concern. I propose replacing mutable classes with immutable ones - starting with things that can actually break, i.e. types / fields / property resolvers etc. I believe those aren't part of the public API, so it won't be a breaking change.

After refactoring the important parts, I would have refactored the rest - i.e. field descriptors, schema factory etc, although this would definitely be a breaking change and is not as important.

Thoughts? As always I'll PR the necessary changes.

oprypkhantc:
Closes #581 

This is still a wip, but I need your feedback to finish it. Don't mind the code style & PHPStan, I'll fix it later.

A couple of notes on the implementation:
 - `Cloneable` is a tiny temporary workaround to allow cloning readonly objects. Once PHP 8.3 is out, it can be refactored to regular `clone + $this->prop = $prop` combination
 - I wanted to avoid creation of `ResolverInterface` instances on each resolve to avoid mutations, so instead went with passing the `object $source` parameter
 - `ResolverInterface::executionSource` is the replacement for `ResolverInterface::getObject()` - it allows getting the object that the field/method is called on. This is used by `#[Security('this.someMethod()')]` and for prefetching data `#[Field(prefetchMethod: 'someMethodOnThis')]`
 - Getters on `QueryFieldDescriptor` and `InputFieldDescriptor` are gone since properties are now readonly

And a lot of questions @oojacoboo:
1. Currently `prefetch` is tied into `graphqlite`, such that it's impossible to implement such a feature in userland: there a hard dependency on it in `QueryField`. What do you think of extracting it into a separate attribute and a parameter middleware? Something like `#[Prefetch('methodName', ['arg1'])]` + `PrefetchParameterMiddleware`? That would simplify the `QueryField` further and allow implementation of similar features in userland, without `graphqlite` having to implement them. I know you're against adding more attributes, but this seems logical. Existing `prefetechMethod` and `prefetechArgs` parameters of `#[Field]` can be deprecated and removed later on.

2. Although `ResolverInterface::executionSource` works, I think it's worthwhile thinking on how to improve that. There are multiple problems currently:
   - checks for `null` in resolvers, because of the `__invoke(object|null $source)` signature forced by `ServiceResolver`, which doesn't need a source. All other resolvers do need one and have ugly runtime checks 
   - `->setCallable()` on `QueryFieldDescriptor` which duplicates the logic of some of the existing resolvers
   - most importantly, input constructor parameters are not passed through middlewares, partly due to issues with resolvers. This single-handedly stops me from using constructor parameters
  
   What do you think of refactoring them too in the scope of 7.x? I have a few ideas on mind.

3. `QueryFieldDescriptor` and `InputFieldDescriptor` are both bloated with logic of selecting the `$originalResolver`, being passed 6 properties `$callable`, `$targetMethodOnSource`, `$targetPropertyOnSource`, `$magicProperty`, `$refMethod` and `$refProperty` solely for that purpose. Instead of doing so we can just accept the `$originalResolver` as a constructor parameter and make respective properties of resolvers `public readonly`, so that in the rare cases it's needed they can just do `$descriptor->originalResolver->methodName`

4. I've only added `readonly` in places where I refactored things, but it's probably worthy adding everywhere, also dropping the getters. Maybe as part of a different PR. Thoughts?


oojacoboo:
1.  Can't we just make `prefetchMethod` accept a callable as well?  What's the benefit of an additional annotation here?
2.  These seem like reasonable improvements.  Let's do a separate PR on them though.
3.  I like it
4. I think locking down the public API is a good idea, or else we could find ourselves dealing with more BC break issues down the road.

oojacoboo:
Of note, on the documentation, we only need to update the current src.  The versioned docs shouldn't be edited since they represent the defined version, which is already locked in functionality.  They only need to be updated for mistakes, not changes to the API.

oojacoboo:
Also, when it comes to defining the public API, I realize there are 2 camps.  With the addition of `readonly` props and typing, we have far greater control over how properties are managed and therefore the accessibility of the property can be used for defining the API, instead of getters/setters.

I realize this may be where we're headed in PHP land, and it may take some time before that's a standard, if ever.  The issue I take with this is that is splits your API into methods and properties and it's not always clear which you should be looking for, especially when using something like autocomplete.  Sometimes you may need a getter to massage the output from a property, making it private, and other times you may not, relying on that property being public.  This leaves the dev wondering and it's a bit chaotic IMO.

In short, I'm not sold on the trend of moving to public properties for defining the API.  It's neat, clean, less code - yes.  But, it can also be confusing.  Further, most of GraphQLite already uses getters/setters, so a move away from this only bifurcates the public API further.

oprypkhantc:
> It sounds like you're trying to reverse the order of processing for prefetching. Instead of defining how a prefetch is handled for a given field, on that field, you're wanting to add a middleware that will prefetch for all fields based on some logic?

Not really, the order and all of the behaviour will stay exactly the same. I'm just not satisfied with the implementation of it. Currently `QueryField` checks whether a prefetch is needed and if so, changes the resolver to return a deferred result. I want to remove any references to `prefetch` inside of `QueryField` and extract them into middlewares. For prefetch, I was thinking of a `PrefetchParameterMiddleware`:
```php
public function mapParameter(ReflectionParameter $parameter, DocBlock $docBlock, Type|null $paramTagType, ParameterAnnotations $parameterAnnotations, ParameterHandlerInterface $parameterMapper): ParameterInterface
    {
        $prefetchAttribute = $parameterAnnotations->getAnnotationByType(Prefetch::class);
        
        if (!$prefetchAttribute) {
            return $parameterMapper->mapParameter($parameter, $docBlock, $paramTagType, $parameterAnnotations);
        }
        
        return new PrefetchDataParameter($prefetchAttribute->method, $prefetchAttribute->args);
    }
```

and a `PrefetchDataParameter`:
```php
public function resolve(object|null $source, array $args, mixed $context, ResolveInfo $info, QueryField $field = null): mixed
    {
        $prefetchBuffer = $context->getPrefetchBuffer($field);
        $prefetchBuffer->register($source, $args); // <--- this is what `QueryField` is currently responsible for

        // QueryField also defers the resolver for prefetch-annotated fields, which could be done on parameter level
        return new Deferred(function () {
            if (! $prefetchBuffer->hasResult($args)) {
                $prefetchResult = $this->computePrefetch($source, $args, $context, $info, $field, $prefetchBuffer);

                $prefetchBuffer->storeResult($prefetchResult, $args);
            }

            return $prefetchResult ?? $prefetchBuffer->getResult($args);
        });
    }
```

That would allow future features that require deferring the parameter resolution possible. And not only in the `graphqlite` itself, but in userland code too.

> On the APIs, public and internal, I think we should stick with getters for defining the API. Going immutable will obviously eliminate setters. We don't need getters for every property internally either. That can be something that's selectively chosen when defining the API.

Got it. I'll revert the getters then on this PR, then submit two more PRs: one for resolvers, the other for `readonly` throughout `graphqlite`.

oojacoboo:
@oprypkhantc not opposed to a `#[Prefetch]` attribute.  I do think the separation of concerns here has a good deal of benefit - I like it.  In general, I do want to see a limited set of attributes on this lib.  I think it's easy to get carried away with adding new functionality through another attribute.  However, I think this one is clear with intent and is deserved.  I like the design on this.  Let's target this on a different PR.

oprypkhantc:
Also while working on `#[Prefetch]` I found an interesting detail: for some bizarre reason, prefetch method is not called statically but rather on the first (random) source instance. This doesn't make sense at all and could lead to problems (like someone accidentally using one of `$this->` properties/methods in the prefetch method).

I believe the more logical solution is to call the method statically when only a method name is specified. Not only will this avoid problems for the users, it'll also simplify the code for graphqlite.

Obviously when a full callable is specified we can just resolve the class from the container instead of calling the method statically. Wdyt?

oojacoboo:
> Regarding `#[Prefetch]`, is removing `#[Field(prefetchMethod: 'some')]` support an OK breaking change or should we first deprecate it?

Are there any issues or limitations in keeping it deprecated?  It'd be ideal to deprecate before removal.

oprypkhantc:
> Are there any issues or limitations in keeping it deprecated? It'd be ideal to deprecate before removal.

No I don't think so; I'll add deprecations then. Also is there a point in supporting the annotation version of `#[Prefetch]`? That wouldn't be difficult either, but I suppose it won't be too long till annotations support is deprecated.

oojacoboo:
Annotation support is already deprecated.  No need to add an annotation.

https://graphqlite.thecodingmachine.io/docs/doctrine-annotations-attributes

oojacoboo:
@oprypkhantc We still need to get that deprecation added.  Is there anything else remaining on this?

oprypkhantc:
@oojacoboo Are you talking about Prefetch-related deprecations? If so, this is included in a different in a different PR, where I refactored Prefetch: https://github.com/thecodingmachine/graphqlite/pull/588/files#diff-44cccfa8f1939494e34146f50de7ef43e88de7e1789ddc53ec4dd4a813bed6c8R67

oprypkhantc:
@oojacoboo Hey. Any chance to review this and after this - the prefetch refactor? There are some other issues I'd like to tackle, but first I want these to get merged.

oojacoboo:
@oprypkhantc this looks great :)

My apologies for the delay in reviewing this.  I'll get the others reviewed as well and look forward to any additional PRs.

Before merging this one, can you confirm that we have adequate test coverage for `prefetch` and the `PrefetchBuffer`?

oojacoboo:
>Should we reorganize tests a bit? Separate fixtures, unit tests and integration tests and add #[Covers] wherever possible? That'd allow actually using phpunit coverage for coverage reliably.

That'd certainly be nice.  It'd give us a little more confidence with these merges.  Everything else looks good though.  I'll go ahead and merge and we can monitor this on master for a bit before tagging a release.

oprypkhantc:
Thanks! It'd also be awesome to include prefetch and resolvers refactor in that release as well.

oojacoboo:
Yes - agreed.

PR #585, opened 2023-03-22
Fixed a typo in example GraphQL query

pvlg:


oojacoboo:
@pvlg I merged this and then realized this PR doesn't target the appropriate docs.  Can you please resubmit this - don't think you can reopen.  The docs change should be on all versions and on the current src, which is used for current and future versions.

PR #588, opened 2023-03-28
Prefetch refactor

oprypkhantc:
This PR targets two things:
- cleanup of `#[Field]`, `QueryFieldDescriptor` and `QueryField` itself, separating `prefetch` functionality from them
- allow userland implementation of prefetch-like features, i.e. deferring the resolution of a field the same way prefetch does

This includes changes from #584, so that would need to be merged first; then this can be reviewed properly. Docs changes are todo.

On the changes:
- `prefetchMethod` parameter of `#[Field]` now trigger a deprecation error
- new `#[Prefetch]` attribute supports callables, both through the container and through static methods: `#[Prefetch([SomeService::class, 'method'])]`, `#[Prefetch([SomeHelper::class, 'staticMethod'])]`, `#[Prefetch('staticMethodInSourceClass')]`
- non static calls on the source object are no longer supported: `#[Prefetch('methodOnSource')]`. There's an exception for the deprecated `#[Field(prefetchMethod: 'method')]` - it continues to work
- as a side effect of moving it to parameter level, multiple `#[Prefetch]` annotations per field are now allowed; order of input parameters in the scheme is determined by order of parameters on the field itself (see `TestTypeWithPrefetchMethods`)


oojacoboo:
@oprypkhantc thanks for this.  I'll get #584 reviewed and hopefully merged soon!

oprypkhantc:
Hey @oojacoboo, this is ready for review. I'll also want to cover `ParameterizedCallableResolver` with tests, but otherwise it's done.

oojacoboo:
Thanks @oprypkhantc, looks good - merged

PR #605, opened 2023-06-20
Make type namespaces available in RootTypeMapperFactoryContext

mdoelker:
Sometimes you need access to the type namespaces to resolve a type name to a class. The enum root type mappers that ship with Graphqlite work like this. Custom root type mappers do not have access to this information currently though, because it is missing in the factory context.

Built-in enum type mappers receive the list of namespaces:
```php
if (interface_exists(UnitEnum::class)) {
    $rootTypeMapper = new EnumTypeMapper($rootTypeMapper, $annotationReader, $symfonyCache, $nsList);
}

if (class_exists(Enum::class)) {
    $rootTypeMapper = new MyCLabsEnumTypeMapper($rootTypeMapper, $annotationReader, $symfonyCache, $nsList);
}
```

`RootTypeMapperFactoryContext` is missing this info though:
```php
$rootSchemaFactoryContext = new RootTypeMapperFactoryContext(
    $annotationReader,
    $typeResolver,
    $namingStrategy,
    $typeRegistry,
    $recursiveTypeMapper,
    $this->container,
    $namespacedCache,
    $this->globTTL,
);
```

I have a custom root type mapper for another enum library and am forced to pass in the type namespaces again:
```php
->addTypeNamespace('My\\Custom\\Ns\\')
->addRootTypeMapperFactory(new MarcMabeEnumTypeMapperFactory(['My\\Custom\\Ns\\']))
```

This PR adds the list of namespaces to the factory context so that root type mapper factories can pass it on to the created mapper instance.

### DIFF
--- a/src/Mappers/Root/RootTypeMapperFactoryContext.php
+++ b/src/Mappers/Root/RootTypeMapperFactoryContext.php
@@ -11,6 +11,7 @@ use TheCodingMachine\GraphQLite\Mappers\RecursiveTypeMapperInterface;
 use TheCodingMachine\GraphQLite\NamingStrategyInterface;
 use TheCodingMachine\GraphQLite\TypeRegistry;
 use TheCodingMachine\GraphQLite\Types\TypeResolver;
+use TheCodingMachine\GraphQLite\Utils\Namespaces\NS;
 
 /**
  * A context class containing a number of classes created on the fly by SchemaFactory.
@@ -18,6 +19,9 @@ use TheCodingMachine\GraphQLite\Types\TypeResolver;
  */
 final class RootTypeMapperFactoryContext
 {
+    /**
+     * @param iterable<NS> $typeNamespaces
+     */
     public function __construct(
         private readonly AnnotationReader $annotationReader,
         private readonly TypeResolver $typeResolver,
@@ -26,6 +30,7 @@ final class RootTypeMapperFactoryContext
         private readonly RecursiveTypeMapperInterface $recursiveTypeMapper,
         private readonly ContainerInterface $container,
         private readonly CacheInterface $cache,
+        private readonly iterable $typeNamespaces,
         private readonly int|null $globTTL,
         private readonly int|null $mapTTL = null,
     ) {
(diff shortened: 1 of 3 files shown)


src/Mappers/Root/RootTypeMapperFactoryContext.php
```diff
@@ -26,6 +30,7 @@ public function __construct(
         private readonly RecursiveTypeMapperInterface $recursiveTypeMapper,
         private readonly ContainerInterface $container,
         private readonly CacheInterface $cache,
+        private readonly iterable $typeNamespaces,
```
oojacoboo, line 33: It'd be nice to have some validation for an iterable here.  Is there any reason why we shouldn't do this?

oojacoboo:
@mdoelker I'd like to not add more BC stuff to this lib.  We need to limit complexity as much as possible for the sake of maintenance.  If you can work with an adapter for this, that's going to be best.  We need to be moving forward.  Migrations to native Enums really shouldn't be difficult.

src/Mappers/Root/RootTypeMapperFactoryContext.php
mdoelker: What do you have in mind? I am not sure I understand what you are referring to.

mdoelker:
@oojacoboo Sure. Getting this PR's change in would make that even easier and allow for a simple plugin-like usage.

src/Mappers/Root/RootTypeMapperFactoryContext.php
oojacoboo: Seeing as `$typeNamespaces` could be `['foo', 1, Bar::class]`, it'd be nice to validate that input to ensure it's proper namespaces.

mdoelker: Right. I don't think that's necessary here as we're dealing with a class that is only ever instantiated from internal code aka the lib's SchemaFactory. We have a couple of other places where we just rely on the iterable type hint and the IDE's extended type checking capabilities. I don't wanna add a runtime check where the statical analysis of the IDE should be sufficient.

oojacoboo:
Thanks for the PR @mdoelker!

PR #606, opened 2023-06-23
Enum descriptions and deprecation support

mdoelker:
This PR adds support for docblock descriptions for native enums and enum cases, as well as deprecation tag support for cases.

This enum
```php
/**
 * This is the description of the enum itself
 */
enum TestEnum
{
    /**
     * This is a case description
     */
    case FOO;

    /**
     * This is another case description
     * @deprecated it should not be used anymore
     */
    case BAR;
}
```

results in the following type description in the schema:
```json
{
  "kind": "ENUM",
  "name": "TestEnum",
  "description": "This is the description of the enum itself",
  "specifiedByUrl": null,
  "fields": null,
  "inputFields": null,
  "interfaces": null,
  "enumValues": [
    {
      "name": "BAR",
      "description": "This is another case description",
      "isDeprecated": true,
      "deprecationReason": "it should not be used anymore"
    },
    {
      "name": "FOO",
      "description": "This is a case description",
      "isDeprecated": false,
      "deprecationReason": null
    }
  ],
  "possibleTypes": null
}
```

GraphiQL output
image
image

### DIFF
--- a/src/Mappers/Root/EnumTypeMapper.php
+++ b/src/Mappers/Root/EnumTypeMapper.php
@@ -10,6 +10,7 @@ use GraphQL\Type\Definition\OutputType;
 use GraphQL\Type\Definition\Type as GraphQLType;
 use MyCLabs\Enum\Enum;
 use phpDocumentor\Reflection\DocBlock;
+use phpDocumentor\Reflection\DocBlockFactory;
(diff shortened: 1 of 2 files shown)


oojacoboo:
@mdoelker thanks for this - looks great!

PR #608, opened 2023-06-27
Doc: Fix UseInputType example as an attribute

cvergne:
Hi,

A very small contribution about the documentation of UseInputType attribute.

The current documentation shows the use of `UseInputType` attribute in the same way than before with annotation by targeting method instead of parameter.

The `for` parameter doesn't exists in the Attribute (_unlike the annotation_), and the whole attribute should be moved just before the targeted parameter.


oojacoboo:
@cvergne thanks for the PR!

PR #612, opened 2023-08-04
Automatic query complexity

### LINKED ISSUE #610: Automatic query complexity
Hey.

There was a similar issue here, see last comment: https://github.com/thecodingmachine/graphqlite/issues/306 . What's important here is the following: it'd be nice to have `graphqlite` automatically compute query complexity using pre-defined settings and an attribute. Some other GraphQL implementations (for other languages) have been doing this for a while and it's pretty neat.

For example, here's a guide from HotChocolate (.NET graphql implementation): https://chillicream.com/docs/hotchocolate/v13/security/operation-complexity

The way they do it is pretty simple: they count every non-object and non-list field as `1`, and every object or list or anything with a custom resolver as `5`, and allow overriding that through an annotation:

```graphql
type Query {
  books(take: Int = 10): [Book] @cost(complexity: 5, multipliers:[take])
}

type Book {
  title
  author: Author @cost(complexity: 5)
}

type Author {
  name
}
```

Given this example, a query like this:
 
```graphql
query {
  books {
    title
  }
}
```

costs `10` (`take` parameter) * (`5` (each Book in `books` cost) + `1` (title of each Book))  = 10 * (5 + 1) = 60

While a query like that:

```graphql
query {
  books {
    title
    author {
      name
    }
  }
}
```

costs 10 * (5 + 1 + 5) = 110.

The implementation consists of a new attribute `#[Cost(int $complexity = 1, ?int $defaultMultiplier = null, array $multipliers = [])]` and a `ComplexityFieldMiddleware`. Again, thanks to `webonyx/graphql-php` supporting query complexity out of the box, it's pretty trivial, yet quite nice.

I've already implemented this in our package and would like to backport this to `graphqlite`, if there's any interest in it here.

oprypkhantc:
Closes #610

This is based on [Hot Chocolate's query complexity](https://chillicream.com/docs/hotchocolate/v13/security/operation-complexity). The way this works is it sets a `complexityFn` on each field, which `webonyx/graphql-php` uses to determine each field's complexity.

I felt like including Cost information in the schema is important so clients can actually estimate a cost for a specific query before even thinking of writing it. Unfortunately, with the current state of GraphQL, there's no other way to add "meta" information to fields other than the description. This is why I'm adding cost information to the field "comments", although I'd much prefer something like a `@cost` directive. Sadly, that's not possible.


src/Annotations/Cost.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TheCodingMachine\GraphQLite\Annotations;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_PROPERTY | Attribute::TARGET_METHOD)]
+class Cost implements MiddlewareAnnotationInterface
```
oprypkhantc, line 10: add @Cost to list of annotations in docs

src/Middlewares/CostFieldMiddleware.php
```diff
@@ -0,0 +1,73 @@
...
+
+            return $cost;
+        };
+
+        return $field;
+    }
+
+    private function buildQueryComment(Cost $costAttribute): string
+    {
+        return 'Cost: ' .
```
oprypkhantc, line 66: add test for this part

oojacoboo (approved):
This is awesome @oprypkhantc - thanks!  I'll be testing an implementation as well.

oojacoboo:
So, one thing I am noticing is that, without the ability to base the multiplier off the actual number of objects retrieved in a collection, you're forced to set your default limits in such a way that's already calculating the upper bounds of the overall complexity limit.  This is mostly a concern if the client isn't always passing the limit (`take`) and it's defaulting `(int $limit = 50)`.  It does make implementation a bit more thoughtful, to achieve optimal complexity calculations.  The benefit is that it encourages more thoughtful default limits and you get better test coverage as it's calculating against the assumed upper bounds.

Regardless, there isn't a way around this to still achieve static analysis - just an observation.

oojacoboo:
@oprypkhantc I don't see anything around mutations.  Will this work on mutations, input type fields?

oprypkhantc:
@oojacoboo It's designed only to work on selections; input fields are not calculated towards query complexity. Is there a reason for it to work for input fields? In our case mutations are generally small, so too many input fields is definitely not a problem for us.

oojacoboo:
The issue with not having cost complexity associated with mutations is that you're left with needing to resolve mutation abuse another way - especially considering that write operations are expensive.  But having costs associated with mutations, you can mitigate this concern.  But adding cost complexity to fields of an input type, you have a more control over the accuracy of the cost calculations for a mutation.

PR #618, opened 2023-09-13
Improved description formatting for cost complexity

oojacoboo:
This is a simple commit, we just add a new line prior to the cost complexity in the description.  I've tested this on multiple GraphQL clients and it seems to be working well.  The current formatting made it difficult to read and the cost complexity is jammed up against existing documentation.

@oprypkhantc let me know if you're aware of any issues where this formatting would be undesirable.


oprypkhantc:
Looks good :)

PR #619, opened 2023-09-18
Fix name of imported type from webonyx for formatter

andrew-demb:
`SerializableError` was a representation of one error, but the formatter returns an array of errors

https://github.com/webonyx/graphql-php/blob/29b5b613bd3ee34f227650f856ee8dd4ce151b88/src/Executor/ExecutionResult.php#L17-L23


andrew-demb:
Sorry, my suggestion was wrong.
Formatter deals with single error.

PR #628, opened 2023-10-15
Use @param tags from constructor with property promotion

oojacoboo:
When using property promotion, particularly with input types, you're required to define the generic type details for lists (array, collection, etc) using a variable docblock.  Not only is this ugly, it requires duplicate docblock definitions for internal docblock usage and is entirely unnecessary.

This PR adds support for constructor docblock generic type definitions with property promotion.

Example:

```php
/**
 * Constructor
 * 
 * @param array<string> $array
 */
#[Input]
public function __construct(
    #[Field]
    public array $array
) {}
```

As opposed to:

```php
#[Input]
public function __construct(
    /** @var array<string> */
    #[Field]
    public array $array
) {}
```

The additional support is fully BC.

### DIFF
--- a/src/Mappers/Parameters/TypeHandler.php
+++ b/src/Mappers/Parameters/TypeHandler.php
@@ -126,22 +126,35 @@ class TypeHandler implements ParameterHandlerInterface
         $varTags = $docBlock->getTagsByName('var');
 
         if (! $varTags) {
-            return null;
-        }
-
-        if (count($varTags) > 1) {
             // If we don't have any @var tags, was this property promoted, and if so, do we have an
             // @param tag on the constructor docblock?  If so, use that for the type.
             if ($refProperty->isPromoted()) {
-                $docBlock = $this->cachedDocBlockFactory->getDocBlock($refProperty->getDeclaringClass()->getConstructor());
+                $refConstructor = $refProperty->getDeclaringClass()->getConstructor();
+                if (! $refConstructor) {
+                    return null;
+                }
+
+                $docBlock = $this->cachedDocBlockFactory->getDocBlock($refConstructor);
                 $paramTags = $docBlock->getTagsByName('param');
                 foreach ($paramTags as $paramTag) {
+                    if (! method_exists($paramTag, 'getVariableName')) {
+                        continue;
+                    }
+
+                    if (! method_exists($paramTag, 'getType')) {
+                        continue;
+                    }
+
                     if ($paramTag->getVariableName() === $refProperty->getName()) {
                         return $paramTag->getType();
                     }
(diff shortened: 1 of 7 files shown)


oojacoboo:
Should be ready.  A review would be nice.

tests/FieldsBuilderTest.php
```diff
@@ -197,6 +200,46 @@ public function test($typeHintInDocBlock)
...
+            PropertyPromotionInputTypeWithoutGenericDoc::class,
+            'PropertyPromotionInputTypeWithoutGenericDocInput',
+        );
+    }
+
+    /**
+     * Tests that the fields builder will properly build an input type using property promotion
+     * with the generic docblock defined on the constructor and not the property directly.
+     *
+     * @requires PHP >= 8.0
```
andrew-demb, line 225:
```suggestion
     * @requires PHP >= 8.0
```

It is not necessary, since library supports only 8.1+

src/Mappers/Parameters/TypeHandler.php
```diff
@@ -124,6 +126,31 @@ private function getDocBlockPropertyType(DocBlock $docBlock, ReflectionProperty
...
+            if ($refProperty->isPromoted()) {
+                $refConstructor = $refProperty->getDeclaringClass()->getConstructor();
+                if (! $refConstructor) {
+                    return null;
+                }
+
+                $docBlock = $this->cachedDocBlockFactory->getDocBlock($refConstructor);
+                $paramTags = $docBlock->getTagsByName('param');
+                foreach ($paramTags as $paramTag) {
+                    if (! method_exists($paramTag, 'getVariableName')) {
```
andrew-demb, line 140:
```suggestion
                    if (! $paramTag instanceof ParamDescriptor) {
```

I propose to check for expected type, not a existed method.

This allows IDE/SA to know about type of returned value from called method and avoid to have dead code in case of removed method in phpdocumentor (if it would not be catched with tests).

https://docs.phpdoc.org/3.0/classes/phpDocumentor-Descriptor-Tag-ParamDescriptor.html

tests/FieldsBuilderTest.php
oojacoboo: Agreed, was following similar fixture formats.  We probably need to clean all that up now that 8.1 is min support.

oojacoboo: Will submit all this under another PR

src/Mappers/Parameters/TypeHandler.php
andrew-demb: Sorry, commented from mobile :)

PR #636, opened 2023-11-18
Allow constructor fields to use middleware

### LINKED ISSUE #565: Support readonly input classes with constructors
Currently, the best you can do when defining an update `#[Input(update: true)]` is to define regular properties without `readonly`:

```php
#[Input(default: true, update: true)]
class UpdateDTO
{
	#[Field] public Layout $layout;
}
```

Then, if you try to
 - mark the class/property as `readonly`, it fails trying to set it: `Cannot initialize readonly property UpdateDTO::$layout from scope TheCodingMachine\\GraphQLite\\Utils\\PropertyAccessor`. This can be fixed by using reflection to set the value instead of setting it directly using `$object->$field = $value;`
 - mark the property as private (in case you instead want to provide a getter for it), it fails again with a similar error
 - lastly, use promoted properties (constructor hydration) - it fails again with `Parameter 'layout' is missing for class 'UpdateDTO' constructor. It should be mapped as required field.`. 

Modern deserialization (Java - [Moshi](https://github.com/square/moshi/blob/master/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.kt#L59), [GSON](https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java#L262)) libraries default to creating an instance without constructor and then hydrating the properties rather than calling the constructor, even if it exists. This makes sense because deserialization usually requires more complex logic than just create an instance, as is the case with `#[Input(update: true)]`. 

The reason I want to use promoted properties is because those inputs classes are used as value objects throughout the project, so it makes perfect sense for them to have a regular constructor. One more thing is unit testing - with a constructor I can do this:
```php
$service->doSomething(new UpdateDTO(layout: Layout::BASIC))
```

So what would sound better to me is always using field-based hydration:

```php
#[Input(default: true, update: true)]
readonly class UpdateDTO
{
	public function __construct() {
		#[Field] public Layout $layout,
	}
}
```

If constructor hydration is deemed important (at the very least for compatibility with older code that _could_ break), we can instead make it opt-in:

```
#[Input(default: true, update: true)]
class UpdateDTO
{
	#[Hydrate]
	public function __construct() {
		#[Field] public Layout $layout,
	}
}
```

Thoughts on using reflection to set the property? And the hydration?

oprypkhantc:
Hey @oojacoboo. This is a draft, but I wanted you to take a look first before I get to fix the tests and polish this. There are no new tests as of now.

This is the minimum fix needed to close #565 and allow constructor fields to have middleware, allowing this kind of code:

```php
#[Input]
class SomeType {
    public function __construct(
        #[Field]
        #[SomeMiddleware]
        public readonly string $field,
    ) {}
}
```

What I did is I refactored resolvers a bit to make sure a single input middleware could be used for both constructor and "set" properties. This is done by requiring resolvers to always return the value that will be set if a parameter is fed into a constructor, as opposed to being set directly through the property.

### DIFF
--- a/src/FailedResolvingInputType.php
+++ b/src/FailedResolvingInputType.php
@@ -10,9 +10,9 @@ use function sprintf;
 
 class FailedResolvingInputType extends RuntimeException
 {
-    public static function createForMissingConstructorParameter(string $class, string $parameter): self
+    public static function createForMissingConstructorParameter(string $original): self
     {
-        return new self(sprintf("Parameter '%s' is missing for class '%s' constructor. It should be mapped as required field.", $parameter, $class));
+        return new self(sprintf("%s. It should be mapped as required field.", $original));
     }
 
     public static function createForDecorator(string $class): self
--- a/src/FieldsBuilder.php
+++ b/src/FieldsBuilder.php
@@ -38,7 +38,13 @@ use TheCodingMachine\GraphQLite\Middlewares\FieldHandlerInterface;
 use TheCodingMachine\GraphQLite\Middlewares\FieldMiddlewareInterface;
 use TheCodingMachine\GraphQLite\Middlewares\InputFieldHandlerInterface;
 use TheCodingMachine\GraphQLite\Middlewares\InputFieldMiddlewareInterface;
+use TheCodingMachine\GraphQLite\Middlewares\MagicPropertyResolver;
 use TheCodingMachine\GraphQLite\Middlewares\MissingMagicGetException;
+use TheCodingMachine\GraphQLite\Middlewares\ServiceResolver;
+use TheCodingMachine\GraphQLite\Middlewares\SourceConstructorParameterResolver;
+use TheCodingMachine\GraphQLite\Middlewares\SourceInputPropertyResolver;
+use TheCodingMachine\GraphQLite\Middlewares\SourceMethodResolver;
+use TheCodingMachine\GraphQLite\Middlewares\SourcePropertyResolver;
 use TheCodingMachine\GraphQLite\Parameters\InputTypeParameterInterface;
 use TheCodingMachine\GraphQLite\Parameters\ParameterInterface;
 use TheCodingMachine\GraphQLite\Parameters\PrefetchDataParameter;
@@ -370,14 +376,6 @@ class FieldsBuilder
                 $type = $this->typeMapper->mapReturnType($refMethod, $docBlockObj);
             }
 
-            $fieldDescriptor = new QueryFieldDescriptor(
-                name: $name,
-                type: $type,
-                comment: trim($description),
-                deprecationReason: $this->getDeprecationReason($docBlockObj),
-                refMethod: $refMethod,
-            );
-
             $parameters = $refMethod->getParameters();
             if ($injectSource === true) {
                 $firstParameter = array_shift($parameters);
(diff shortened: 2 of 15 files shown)


oojacoboo:
@oprypkhantc I'm wondering why you've gone with a middleware on a field?  Why would we not define a global hydration middleware and let that be it?  If there is a need to have field level middleware, that's something that can be implemented via the global middleware in userland.

I haven't had time to look through the code in detail, as I'm currently out of town.  But, I'm assuming it's strictly using reflection now?  Can you detail the supported variations for hydration?  Is the constructor actually called?

oprypkhantc:
> @oprypkhantc I'm wondering why you've gone with a middleware on a field? Why would we not define a global hydration middleware and let that be it? If there is a need to have field level middleware, that's something that can be implemented via the global middleware in userland. Does `SomeMiddleware` implement an interface?
> 
> I haven't had time to look through the code in detail, as I'm currently out of town. But, I'm assuming it's strictly using reflection now? Can you detail the supported variations for hydration? Is the constructor actually called?

With "middleware" I was referring to the existing `InputFieldMiddlewareInterface` that we have for modifying input fields. `SomeMiddleware` is just a regular field middleware annotation implementing `MiddlewareAnnotationInterface`. Previously, you could not use middleware for fields hydrated through the constructor; now you can.

The hydration is unchanged: any fields that are also defined as constructor parameters are passed into the constructor, the rest is set directly by setting the property publicly, no reflection.

oojacoboo:
Okay, I see.  How does this address the rest of the concerns discussed on #565?

oprypkhantc:
It doesn't address most of them. I tried to approach all of the concerns raised in the issue, but so far I've decided that some points probably don't make sense, while others are harder to implement than initially anticipated.

Points from that discussion:
  - hydrating readonly properties (except for constructor promoted) is unchanged and results in an error. This change is non-controversial and relatively trivial to implement, but will need some work on resolvers. Not part of this PR.
  - marking property private (again except for constructor) won't work. There are comments in the code saying this is intentionally not supported, so I believe we should leave it at that, at least for now. 
  - constructor hydration will now work correctly with all the middleware. It still won't work with the "update: true" flag of input types, but there's nothing that can be done here
  - treating `null` value and "optional" field (i.e. missing key in the request) as separate concepts again I didn't do as we've decided it was out of scope. I implemented that separately in a different library.
  - lastly, the custom hydration. I attempted to implement that, but quickly realized I don't actually know why we even need this :) One of the cases that custom hydration middleware was meant to solve was proper constructor hydration, and it is now solved with this PR. I'm not aware of any other use cases, hence it's hard to predict what needs (or doesn't need) to be built here. It sounds good in theory, but I'd rather wait and gather other people's concerns and only then implement something

oojacoboo:
@oprypkhantc this PR seems fine.  I was confused though, because it doesn't close #565, and really doesn't address the main concerns of that ticket.

Custom hydration middleware offers the ability for someone to implement any logic deemed necessary, prior to, or as part of, hydration of an input type.  Validation is a clear example of why someone might want to do this.  But surely there are other possible reasons.

A custom middleware isn't really that difficult.  Basically we'd ship the default hydrator that'd be sufficient for most use cases, only check the config to see if a custom one has been provided and use that instead.

If we move to using reflection for constructor properties, to get around the private and readonly limitations, and don't call the constructor directly, if someone does want to execute the constructor as part of hydration, they'd have to resort to a custom hydrator.  This allows for us to solve the issues presented in #565, while still providing some solution for BC support.

oprypkhantc:
> @oprypkhantc this PR seems fine. I was confused though, because it doesn't close #565, and really doesn't address the main concerns of that ticket.

Sorry for the confusion 🥲 

> A custom middleware isn't really that difficult. Basically we'd ship the default hydrator that'd be sufficient for most use cases, only check the config to see if a custom one has been provided and use that instead.

Unfortunately it's not as easy as that. As of right now, input field resolvers set the property directly or call a setter method. This means that if we do ship some sort of hydration customization, it won't be possible to overwrite literally half of the process - you either call a resolver and the property gets set by `graphqlite`, or you don't call it and then you don't get the value you need.

To me that's not a good or a complete solution. And to fix that, input resolvers shouldn't be setting the property directly, but rather just returning the value so it can be passed through to the hydrator, which in turn can do anything with it. But the problem here is that now the hydrator must also know which fields are set directly through properties, which fields have setters associated with them and which fields are magic properties.

All this complicates the hydration a lot, so I didn't want to deal with all that as well.

> If we move to using reflection for constructor properties, to get around the private and readonly limitations, and don't call the constructor directly, if someone does want to execute the constructor as part of hydration, they'd have to resort to a custom hydrator. This allows for us to solve the issues presented in #565, while still providing some solution for BC support.

Since creating that issue I actually changed my mind regarding constructor hydration. I no longer want hydration to omit the constructor, which is why I'm PRing these changes - to allow using the constructor at all. Of course having custom hydration support is always better than not having it; but I believe the existing hydration (given this PR and fixes to setting readonly/private properties) is sufficient to satisfy 99% users, including me.

Issue #336, opened 2021-04-15
graphql api returns 500 http codes on unknown types


massimilianobraglia:
Hey there,

```graphql
query foobar($foobar: Foobar!){
  foobar(foobar: $foobar){
    foobar
  }
}
```

Calling this query on an environment that has no `Foobar` type defined results in a 500 http response like in this screenshot.

Is it normal? Shouldn't it be a 400 response?

![image](https://user-images.githubusercontent.com/17158942/114885983-7633d480-9e07-11eb-9486-f8ac33a89723.png)

oojacoboo:
Yes.  I think there are a few cases where the http response codes aren't being properly set.  Care to submit a pull request?

massimilianobraglia:
I'll try to find some time for this one! Not sure when :/

oojacoboo:
@massimilianobraglia be a great time to get a PR merged in.  Thanks.

sebcar:
I'll try to solve this issue. @oojacoboo any suggestions about the test case?

oojacoboo:
@sebcar I think an example like the one mentioned should be sufficient here.  There are a few other cases as well, where GQL syntax results in a 500.  I'm not sure of all these cases though.  GraphQLite uses Webonyx's GQL parser under the hood.  I don't think we'll be able to wrap that in a try/catch either, unless there is a specific Exception being thrown for these parsing errors.

sebcar:
I can't find a way in the test case to get the execution result for CannotMapTypeException

oojacoboo:
@sebcar If there is an exception being thrown, why would there be a result?  Presumably, execution was halted before there was a result.

sebcar:
@oojacoboo that is true. I was trying to use the HttpCodeDeciderTest

PR #646, opened 2024-01-03
fix: RecursiveTypeMapper returns exception which can't be handled by StandardServer

fogrye:
RecursiveTypeMapper throws CannotMapTypeException when fails to find class by name. This exception goes through the GraphQl executor instead of catching and wrapping into ExecutionResult. Non wrapped exception does not trigger HttpCodeDecider. Schema test trust more to input than configuration which is not optimal, great possibility to improve configuration validation.

Fixes: #336

### DIFF
--- a/src/Mappers/RecursiveTypeMapper.php
+++ b/src/Mappers/RecursiveTypeMapper.php
@@ -4,12 +4,14 @@ declare(strict_types=1);
 
 namespace TheCodingMachine\GraphQLite\Mappers;
 
+use GraphQL\Error\Error;
 use GraphQL\Type\Definition\InputObjectType;
 use GraphQL\Type\Definition\InputType;
 use GraphQL\Type\Definition\InterfaceType;
 use GraphQL\Type\Definition\NamedType;
 use GraphQL\Type\Definition\OutputType;
 use GraphQL\Type\Definition\Type;
+use GraphQL\Utils\Utils;
 use Psr\SimpleCache\CacheInterface;
 use ReflectionClass;
 use RuntimeException;
@@ -520,7 +522,6 @@ class RecursiveTypeMapper implements RecursiveTypeMapperInterface
 
             return $this->mapClassToInterfaceOrType($className, null);
         }
-
-        throw CannotMapTypeException::createForName($typeName);
+        throw Error::createLocatedError("Unknown type ".Utils::printSafe($typeName));
     }
 }
--- a/src/Mappers/RecursiveTypeMapper.php
+++ b/src/Mappers/RecursiveTypeMapper.php
@@ -520,6 +520,7 @@ class RecursiveTypeMapper implements RecursiveTypeMapperInterface
 
             return $this->mapClassToInterfaceOrType($className, null);
         }
+
         throw TypeNotFoundException::createError($typeName);
     }
 }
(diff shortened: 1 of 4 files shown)


oojacoboo:
Thanks for this @fogrye.  That skipped test - can we get that resolved?  What's the issue with it?  I'm not seeing where it's using a request in that test either.

fogrye:
@oojacoboo it meant to fail when test query request entity which isn't presented because of bad config but after fix stoped. Now `doTestSchema` throws unexpected exception because `$result->toArray()` returns array without `data` key but with `errors`. 

I see there only option for rewrite: check that with bad config (like not existing folder) nothing happens or new exception been thrown.

oojacoboo:
Can we still use `CannotMapTypeException` and set the `Error` from the `createLocatedError` as the previous exception value?  Maybe by adding a new method like `CannotMapTypeException::createFromLocatedError`?

fogrye:
@oojacoboo unfortunately no, `GraphQL\Server\Helper` is too strict about types.

oojacoboo:
Why can't/shouldn't `CannotMapTypeException` just extend `GraphQL\Error\Error`?

fogrye:
@oojacoboo I see the point in keeping everything clean and consistent but here is some problems:
1. If `CannotMapTypeException` becomes a child of `Error` all  errors it constructs becomes "soft" and handled which is a big change.
2. All exceptions in `CannotMapTypeException` some way referred to dev's mistake while in case of `RecursiveTypeMapper` it's end of execution which deserves some "special attention".
3. Also `RecursiveTypeMapper` has nothing to take from it's possible parent and this extend just a bad inheritance.

Of course it's up to you to as I'm newbie here but I would wish to clean up mappings a little including rethinking about how they trigger errors. Like, as an example, currently there is no way to throw error with reference to place in gql request because mapper which triggers them know nothing about which node it's processing.

oojacoboo:
Good points @fogrye.  Let's find a way to preserve that test.  Maybe it makes sense to create a new Exception class for this scenario, then we can expect that Exception in that test commented out?

I agree that some improvements could be made on the mappings.  If you want to submit a PR that takes a shot at this, that'd certainly be appreciated.

oojacoboo:
@fogrye just a reminder on this.  It'd be nice to get this merged in before it gets stale.  We need tests though.

fogrye:
@oojacoboo had a busy time lately, will try to make it for tests soon.

tests/SchemaFactoryTest.php
```diff
@@ -168,9 +169,8 @@ public function testException2(): void
         $factory->createSchema();
     }
 
-    private function doTestSchema(Schema $schema): void
+    private function _doTestSchema(Schema $schema): ExecutionResult
```
oojacoboo, line 172: Let's not name private methods with a leading underscore, this is a really outdated code style that's really no longer used.

src/Mappers/RecursiveTypeMapper.php
```diff
@@ -520,7 +520,6 @@ public function mapNameToType(string $typeName): Type&NamedType
 
             return $this->mapClassToInterfaceOrType($className, null);
         }
-
-        throw CannotMapTypeException::createForName($typeName);
+        throw TypeNotFoundException::createError($typeName);
```
oojacoboo, line 523: An extra line before a throw statement is a good codestyle, much like a line before a return statement.

oojacoboo (changes requested):

oojacoboo:
@fogrye thanks for this - couple small code style changes and I think it's ready to merge :)

oojacoboo (approved):

PR #649, opened 2024-01-27
Support subscription operations

oojacoboo:
Currently, GraphQLite doesn't support `subscription` operations at all.  However, basic support for the GQL syntax is supported in webonyx.  This PR provides basic support for subscriptions.

By design, this subscription implementation is basic.  After much research and discussion, it's apparent that subscriptions will need to be implemented in a large variety of ways.  This implementation is non-opinionated, only replicating the same implementation style as queries and mutations.

What it does, is allow the subscription GQL syntax, routing to the new `#[Subscription]` annotated controller method.  The behavior will be the same as a mutation or query - no difference there.

Future improvements could be made to this to support long-running PHP processes.  I'm thinking the controller method could return a generator/fiber.


oprypkhantc:
Cool, good seeing some work done in that direction :)

PR #662, opened 2024-03-14
Add a comment mentioning that FilesystemCache can be used when APCu isn't available

shish:
Add a comment mentioning that FilesystemCache can be used when APCu isn't available

APCu isn't enabled by default, which makes the example crash, and there are no hints as to what the user is supposed to do about it


oojacoboo:
I think this one could be updated to be a code comment.  But APCu should be encouraged for performance purposes.  People shouldn't assume that the filesystem is suitable for cache.  It really shouldn't be used.

So, it should stay as APCu with a comment that the Filesystem can be used for testing, but isn't encouraged for performance reasons!

Issue #553, opened 2023-02-03
"smallest working example using no framework" doesn't work


shish:
```
// $cache is a PSR-16 compatible cache.
// $container is a PSR-11 compatible container.
$factory = new SchemaFactory($cache, $container);
```

$cache and $container aren't defined anywhere in the example code :(

I've created an empty in-memory cache object to be used as $cache, and an empty container object to be used as $container, and then my schema comes out empty.

I'm _guessing_ that _somehow_ I'm supposed to be supplying a `$container` that is pre-populated full of `Types`? (Does `addTypeNamespace` not automatically load all the types from the namespace?)

oojacoboo:
Good question.  The service container is certainly used for instantiation of various services/classes.  Your types, operations and factories should be callable through your container.  So yes, "pre-populated", if you'd like to say it like that.  Although most containers don't actually "pre-populate", but instead handle instantiation and caching.  If they weren't called in this fashion, you wouldn't be able to make use of your container service.

shish:
So I'm trying to reverse-engineer graphqlite's expectations by providing a PSR-11 compatible container which just logs calls:

```php
class C implements ContainerInterface
{
    public function get(string $id): mixed {
        echo("get($id)\n");
    }
    public function has(string $id): bool {
        echo("has($id)\n");
        return false;
    }
}
```

... but apparently it never gets called at all?

Then I wonder if I'm doing something wrong somewhere else:

```php
<?php

namespace Shimmie2;

require_once("vendor/autoload.php");

use TheCodingMachine\GraphQLite\SchemaFactory;
use TheCodingMachine\GraphQLite\Annotations\Type;
use TheCodingMachine\GraphQLite\Annotations\Field;
use TheCodingMachine\GraphQLite\Annotations\Query;
use TheCodingMachine\GraphQLite\Annotations\Mutation;
use GraphQL\Utils\SchemaPrinter;

#[Type]
class TagUsage
{
    #[Field]
    public string $tag;
    #[Field]
    public int $uses;

    /**
     * @return TagUsage[]
     */
    #[Query]
    public static function tags(string $search, int $limit=10): array {
        return []; // get from database
    }
}

class C implements \Psr\Container\ContainerInterface
{
    public function get(string $id): mixed {
        echo("get($id)\n");
    }
    public function has(string $id): bool {
        echo("has($id)\n");
        return false;
    }
}

$cache = new \Sabre\Cache\Memory();
$container = new C();
[...]

PR #663, opened 2024-03-14
Make the "smallest working example" actually work, see #553

shish:
Make the "smallest working example" actually work, see #553

This example had implicit-and-undocumented dependencies on composer.json, and explicit-but-unexplained dependencies on "a container" and "a cache", leaving the installation and configuring of these as an exercise to the reader. I have changed it so that the example code can be copy-pasted verbatim, and the user will end up with an example which runs, which I find much easier to learn from :)

I'm still not actually sure if I'm doing this correctly, because #553 mentioned loading classes via the PSR-4 autoloader, and I found that it only worked when I manually listed out my controller classes inside the PSR-11 container, but it works...

### DIFF
--- a/website/docs/other-frameworks.mdx
+++ b/website/docs/other-frameworks.mdx
@@ -102,15 +102,63 @@ $result = GraphQL::executeQuery($schema, $query, null, new Context(), $variableV
 
 The smallest working example using no framework is:
 
-```php
+```json title="composer.json"
+{
+  "autoload": {
+    "psr-4": {
+      "App\\": "src/"
+    }
+  },
+  "require": {
+    "thecodingmachine/graphqlite": "^6",
+    "mouf/picotainer": "^1.1",
+    "symfony/cache": "^4.2"
+  }
+}
+```
+
+```php title="src/Controllers/MyController.php"
+<?php
+namespace App\Controllers;
+
+use TheCodingMachine\GraphQLite\Annotations\Query;
+
+class MyController
+{
+    #[Query]
+    public function hello(string $name): string
+    {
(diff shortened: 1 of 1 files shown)


oojacoboo:
Thanks for this @shish!

Let's stick with APCu in the example with a code comment for Filesystem, like #662.

website/docs/other-frameworks.mdx
```diff
@@ -102,15 +102,63 @@ $result = GraphQL::executeQuery($schema, $query, null, new Context(), $variableV
...
-```php
+```json title="composer.json"
+{
+  "autoload": {
+    "psr-4": {
+      "App\\": "src/"
+    }
+  },
+  "require": {
+    "thecodingmachine/graphqlite": "^6",
```
oojacoboo, line 113: I think we can use an asterisk, `*` for the version here.  We're now at 7.0 - no need to specify.

oojacoboo (changes requested):

shish:
(I do wonder if it'd be worth having example code as actual code (either in an "examples" folder, or a "graphqlite-examples" respository), rather than random code snippets inside a markdown text file -- that way we can use eg github actions to run the code and ensure that it both works and stays working 🤔 )

oojacoboo:
Let's update this PR such that the docs direct the reader to the examples, instead of maintaining any code here.

oojacoboo:
> Let's update this PR such that the docs direct the reader to the examples, instead of maintaining any code here.

@shish ^^

oojacoboo:
@shish would be nice to get this merged.  If you're giving up, I'll have to close it.

PR #667, opened 2024-03-22
Use ComposerFinder's useAutoloading method

### LINKED ISSUE #659: GraphQLite attempts to load non-class files, resulting in weird errors
Hey.

We have a `modules/` folder that maps to `Modules\` PSR-4 namespace. This namespace was also added to be globbed by graphqlite using `->addTypeNamespace('Modules\\')` and `-> addControllerNamespace('Modules\\')`. That namespace also contains non-class files (such as lang files and others), as well as pre-8.0 enums that require autoloading before they can be used.

If you try to blindly require these files, you'll get errors as those non-class files are meant to be loaded from a specific context, and those enums require a special autoloader to call `::initialize()` static method on them. 

The problem is that GraphQLite does exactly that - blindly load all files using `require_once` that appear as classes to it. Obviously, this is not desired, as it forces you to list every class separately in calls to `->addTypeNamespace()` and `->addControllerNamespace()` to avoid loading what it can't load.

This is happening because of this fix: https://github.com/thecodingmachine/graphqlite/blob/master/src/Utils/Namespaces/NS.php#L55. It was done specifically to avoid issues with autoloading, and to allow autoloading incorrectly namespaced classes.

Instead of doing that, GraphQLite can instead attempt to load a class through autoloading and simply skip it if it fails. This will break loading of classes that have incorrect namespaces.

What do you think @oojacoboo? I'll draft a PR if it's an acceptable BC.

oprypkhantc:
Closes #659

This implementation is much simpler: now that we use kcs/class-finder and that it has autoloading enabled by default, the only thing left to do is provide an option to disable it if you need the "legacy" behaviour for whatever reason.

I've tested the new autoloading flow in our project and it's behaving exactly as expected.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "symfony/expression-language": "^4 || ^5 || ^6 || ^7",
         "thecodingmachine/cache-utils": "^1",
         "webonyx/graphql-php": "^v15.0",
-        "kcs/class-finder": "^0.4.0"
+        "kcs/class-finder": "dev-master"
     },
     "require-dev": {
         "beberlei/porpaginas": "^1.2 || ^2.0",
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "symfony/expression-language": "^4 || ^5 || ^6 || ^7",
         "thecodingmachine/cache-utils": "^1",
         "webonyx/graphql-php": "^v15.0",
-        "kcs/class-finder": "dev-master"
+        "kcs/class-finder": "^0.5.0"
     },
     "require-dev": {
         "beberlei/porpaginas": "^1.2 || ^2.0",
--- a/src/GlobControllerQueryProvider.php
+++ b/src/GlobControllerQueryProvider.php
@@ -6,7 +6,6 @@ namespace TheCodingMachine\GraphQLite;
 
 use GraphQL\Type\Definition\FieldDefinition;
 use InvalidArgumentException;
-use Kcs\ClassFinder\Finder\ComposerFinder;
 use Kcs\ClassFinder\Finder\FinderInterface;
 use Psr\Container\ContainerInterface;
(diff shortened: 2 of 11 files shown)


oojacoboo:
@oprypkhantc Can't you just set your own customized `Finder` instance on the `SchemaFactory`, instead of us modifying it to support this edge case?

```php
$finder = (new ComposerFinder)->useAutoloading(false);
$schemaFactory->setFinder($finder);
```

oprypkhantc:
@oojacoboo This PR is mostly for backwards compatibility. Let me explain:

Up until now, graphqlite used `include_once` to load classes - both before the migration to `kcs/class-finder` and after. The most recent version of `kcs/class-finder` at the moment is 0.4.0, which still uses `include_once` by default, effectively meaning `useAutoloading(false)`.

Now the author of the library pushed a commit to `master` changing the default to use autoloading instead, meaning once we upgrade to `0.5.0`, the behaviour will change for all users of graphqlite. It also makes some tests in graphqlite fail.

With this PR, I've upgraded to `dev-master` (0.5.0 as soon as the release is out) and added an autoloading opt-out option for those who need to use the legacy `include_once` behaviour for whatever reason. I can remove this option of course and I'll gladly do so :)

We'll still need this PR to fix the tests and add a `kcs/class-finder ^0.5.0` constraint though :)

oojacoboo:
@oprypkhantc yep - got it.

I think we remove this option in GraphQLite and update documentation on how to disable using something like the example code I provided above.

oprypkhantc:
@oojacoboo Done :) Added a small section in the docs about autoloading. Also I'm not sure why docs generation fails on CI - it builds locally no problem 👀

oojacoboo:
Merged.  Sometimes the test runners just crap out and need to be re-run.

PR #668, opened 2024-03-22
Add automatically-verified working examples

shish:
Add "no framework" example to examples folder

Having an example as actual code rather than as out-of-context snippets inside a text file means that we can run the code, and ensure that it actually works

Submitting mostly for discussion to see if people like this idea

TODO:
- have a github action which runs the code and ensures that the "hello world" query works (thus ensuring that the example code is correct)
- install graphqlite from parent folder instead of downloading the published 6.X (thus ensuring that the example is always in-sync with the library it is an example of)
- update the documentation to reference the files in this repository, rather than embedding hard-coded out-of-context snippets into the .mdx files (thus ensuring that the documentation stays in-sync with the known-good example code)

### DIFF
--- a/examples/no-framework/index.php
+++ b/examples/no-framework/index.php
@@ -1,0 +1,41 @@
+<?php
+use GraphQL\GraphQL;
+use GraphQL\Type\Schema;
+use TheCodingMachine\GraphQLite\SchemaFactory;
+use TheCodingMachine\GraphQLite\Context\Context;
+
+use Symfony\Component\Cache\Simple\FilesystemCache;
+use Mouf\Picotainer\Picotainer;
+use GraphQL\Utils\SchemaPrinter;
+use App\Controllers\MyController;
+
+require_once __DIR__ . '/vendor/autoload.php';
+
+// $cache is any PSR-16 compatible cache.
+$cache = new FilesystemCache();
+
+// $container is any PSR-11 compatible container which has
+// been populated with your controller classes.
+$container = new Picotainer([
+        MyController::class => function() {
+                return new MyController();
+        },
+]);
+
+$factory = new SchemaFactory($cache, $container);
+$factory->addControllerNamespace('App\\Controllers')
+        ->addTypeNamespace('App');
+
+$schema = $factory->createSchema();
+
+$rawInput = file_get_contents('php://input');
+$input = json_decode($rawInput, true);
+$query = $input['query'];
+$variableValues = isset($input['variables']) ? $input['variables'] : null;
+
+$result = GraphQL::executeQuery($schema, $query, null, new Context(), $variableValues);
+$output = $result->toArray();
+
+header('Content-Type: application/json');
+echo json_encode($output) . "\n";
+
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -89,3 +89,33 @@ jobs:
           fail_ci_if_error: false # optional (default = false) - Need CODECOV_TOKEN
         # Do not upload in forks, and only on php8.3, latest deps
         if: ${{ github.repository == 'thecodingmachine/graphqlite' && matrix.php-version == '8.3' && matrix.install-args == '' }}
+
+  examples:
+    name: Check Examples
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        example: ['no-framework']
+      fail-fast: false
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v4"
+      - name: "Install PHP with extensions"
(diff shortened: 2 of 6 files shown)


oojacoboo:
@shish I think this is a reasonable solution.  However, it's not going to remain accurate unless there are tests for it.

We should probably update the `.github/workflows/continuous_integration.yml` to include a step that can execute each of the examples and assert the results.

Another alternative would be adding a PHPUnit test that executes the examples.

examples/no-framework/composer.lock
oojacoboo, line 1: Let's not commit the lock file.  Add this to a `.gitignore`.

oojacoboo (changes requested):

oojacoboo:
Converted this to a draft until the TODO items are completed.

shish:
For reasons I am struggling to figure out, this runs fine on my dev machine but fails in github T_T

```
cd examples/no-framework
composer install --no-interaction --no-progress --prefer-dist
php -S localhost:8080 &
sleep 3
curl --silent -X POST -H "Content-Type: application/json" \
            -d '{"query":"{ hello(name: \"World\") }"}' \
            http://localhost:8080/graphql
```

Local output:
```
{"data":{"hello":"Hello World"}}
```

Output when run on github:
```
{"errors":[{"message":"Cannot query field \"hello\" on type \"Query\".","locations":[{"line":1,"column":3}]}]}
```

Interestingly when I run the published graphqlite@6.x it works and graphqlite@7.x gives the same error as we see on github...

examples/no-framework/index.php
```diff
@@ -0,0 +1,41 @@
...
+// $container is any PSR-11 compatible container which has
+// been populated with your controller classes.
+$container = new Picotainer([
+        MyController::class => function() {
+                return new MyController();
+        },
+]);
+
+$factory = new SchemaFactory($cache, $container);
+$factory->addControllerNamespace('App\\Controllers\\')
```
oojacoboo, line 26: So, I don't think the trailing slash is needed on these namespaces.  I'm unsure if that's causing the issue or not.

oojacoboo:
> For reasons I am struggling to figure out, this runs fine on my dev machine but fails in github T_T
> 
> ```
> cd examples/no-framework
> composer install --no-interaction --no-progress --prefer-dist
> php -S localhost:8080 &
> sleep 3
> curl --silent -X POST -H "Content-Type: application/json" \
>             -d '{"query":"{ hello(name: \"World\") }"}' \
>             http://localhost:8080/graphql
> ```
> 
> Local output:
> 
> ```
> {"data":{"hello":"Hello World"}}
> ```
> 
> Output when run on github:
> 
> ```
> {"errors":[{"message":"Cannot query field \"hello\" on type \"Query\".","locations":[{"line":1,"column":3}]}]}
> ```
> 
> Interestingly when I run the published [graphqlite@6.x](mailto:graphqlite@6.x) it works and [graphqlite@7.x](mailto:graphqlite@7.x) gives the same error as we see on github...

Version 7 uses https://github.com/alekitto/class-finder, which will look to the Composer autoload file for namespaces, by default.

shish:
Any idea what needs to change in the code to make it work? If you download this PR and run the test steps from the github action, does it work for you? Any idea why it works on my laptop but not on github?

FYI, I don't have a "known-good" graphqlite setup that I can compare against, so I'm still not sure if my example code is even _supposed_ to work the way it has been written - is it my code that's broken or something wrong with the framework? Without having a working example that I can compare against, I am debugging blind and I don't even know what direction I'm supposed to be going in 😓 

(I'm not actually a graphqlite user, just somebody who found the docs so frustrating that it was easier to write my own schema-generator from scratch -- I would like to deprecate my own and use graphqlite, but I can't do that if I can't figure out why graphqlite isn't working, and it's hard to figure out why it's not working if there isn't a working example to compare against 😅 )

examples/no-framework/index.php
oojacoboo: I've pushed an update to docs on this already.

oojacoboo:
@shish see #679 - might be the issue you're having?

shish:
> @shish see https://github.com/thecodingmachine/graphqlite/issues/679 - might be the issue you're having?

So that fixed the no-framework example, but it didn't fix the psr-15 example, which is having the same error T_T


> Also, it'd be nice if we could merge all these PRs you have together. That'd help with getting this merged.

The "automatically test examples as part of github actions" and "the no-framework example" are now working. Since the psr-15 example is broken for seemingly unrelated examples, I wonder if we could merge _this_ PR as-is, and then rebase the other PR on top of it to try and debug what's wrong with that one?

PR #669, opened 2024-03-22
Add psr-15 example to examples folder

shish:
Add psr-15 example to examples folder

This example is broken and I don't know why (maybe somebody who knows what they are doing can explain?). But at least now we can _see_ that it is broken, which is better than just silently misleading people...

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -95,7 +95,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        example: ['no-framework']
+        example: ['no-framework', 'psr-15']
       fail-fast: false
     steps:
       - name: "Checkout"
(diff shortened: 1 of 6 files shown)


examples/psr-15/composer.lock
oojacoboo, line 1: Don't commit the lock file - add to `.gitignore`

oojacoboo (changes requested):

shish:
So this is failing and I don't know why :( I tried the same fix which worked for the no-framework example, but no luck

```
$ cd examples/psr-15
$ composer install
$ php -S 127.0.0.1:8080
```

```
$ curl -X POST -d '{"query":"{ hello(name: \"World\") }"}' -H "Content-Type: application/json" http://localhost:8080/graphql
{"errors":[{"message":"Cannot query field \"hello\" on type \"Query\".","locations":[{"line":1,"column":3}]}]}%  
```

oojacoboo:
Closing, as there doesn't seem to be any additional activity here.

Issue #675, opened 2024-04-01
Time to drop annotations


fogrye:
### Discussed in https://github.com/thecodingmachine/graphqlite/discussions/673

<div type='discussions-op-text'>

<sup>Originally posted by **fogrye** March 29, 2024</sup>
Doctrine annotations was deprecated since php 8.0 release, should we drop support of them as php7 is not supported and there are no other reason not to use attributes?</div>
Steps to do:

- rm attributes from fixtures
- polish annotation classes
- rm lib from dependencies

I will create MR for this soon

oojacoboo:
Looking forward to it @fogrye

PR #677, opened 2024-04-07
patch: drop DoctrineAnnotations

fogrye:
BC: AnnotationReader constructor changed and it's methods stops triggering exceptions

TestControllerWithInvalidParameterAnnotation deleted as `for` is ignored when `HideParameter` used as parameter attribute

Also attribute exceptions and classes polished

Fixes: #675


oojacoboo (changes requested):
Thanks again for this PR!  I've finished reviewing it - just some small changes.

fogrye:
@oojacoboo I marked this PR as draft because I though it would be nice to fix documentation with it (and remove php7 examples as we do not support it). Or I can have make that in separate PR. What is better for you?

oojacoboo:
> @oojacoboo I marked this PR as draft because I though it would be nice to fix documentation with it (and remove php7 examples as we do not support it). Or I can have make that in separate PR. What is better for you?

Agreed.  Let's update the docs in this PR.

oojacoboo:
@fogrye thanks for the docs updates.  I think we should go ahead and update the [Annotations Reference](https://graphqlite.thecodingmachine.io/docs/annotations-reference) for attributes and drop the "annotations" name.  So the list of attributes are no longer `@whatever`, but `#[whatever]` or `whatever`.

oojacoboo (approved):

PR #685, opened 2024-05-09
[Error handling] Remove category from exceptions

### LINKED ISSUE #683: [Error handling] Category missing in error response
Hi,

Since the update to **webonyx/graphql-php v15**, the `category` property is missing from `extensions` node in response.

The lib has removed the `category` node ( see [upgrade guide v14 -> v15](https://github.com/webonyx/graphql-php/blob/master/UPGRADE.md#breaking-removed-error-extension-field-category)), which means documentation shows wrong examples and Exceptions in Graphqlite library still have the category getter/setter.

What should be done about that ?
Manually adding back the category in `WebonyxErrorHandler` or totally removing it everywhere ?

cvergne:
As mentioned in #683, category support has already been removed from graphQL error responses with graphqlite because of graphql-php update to v15 ( see [upgrade guide v14 -> v15](https://github.com/webonyx/graphql-php/blob/master/UPGRADE.md?rgh-link-date=2024-04-24T14%3A18%3A16Z#breaking-removed-error-extension-field-category) ), which means `category` is already missing from `extensions` in responses.

The purpose of this PR is to remove the category argument from Exceptions.

> [!CAUTION]
> **BC Break**

The `category` argument has been removed from `GraphQLException`, which means any existing instance of a GraphQLException passing category and/or extensions will trigger an error.

**BEFORE**
```php
throw new GraphQLException('foo', 0, null, 'MyCategory', ['field' => 'foo']);
```

**AFTER**
```php
throw new GraphQLException('foo', 0, null, ['field' => 'foo']);
```

> [!NOTE]
> If you want to keep `category`, you'll have two options : 
> 1. Creating your own `GraphQLException` class and using your own `errorFormatter` method
> 2. Moving the category directly into `extensions`

---

Fixes #683

### DIFF
--- a/src/Exceptions/GraphQLException.php
+++ b/src/Exceptions/GraphQLException.php
@@ -14,7 +14,6 @@ class GraphQLException extends Exception implements GraphQLExceptionInterface
         string $message,
         int $code = 0,
         Throwable|null $previous = null,
-        protected string $category = 'Exception',
         protected array $extensions = [],
     ) {
         parent::__construct($message, $code, $previous);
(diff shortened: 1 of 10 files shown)


oojacoboo (approved):

oojacoboo:
@cvergne thanks for this PR!

PR #695, opened 2024-07-15
chore: remove dependency on thecodingmachine/cache-utils

### LINKED ISSUE #693: Update cache-utils library
Hey, guys. I'm pretty sure I'm writing to wrong repository, but there is the issue with caching library, that no one care in that repo.
I'm just trying to speed up the process)

https://github.com/thecodingmachine/cache-utils/pull/8

Thank you!

xyng:
The package seems to be unmaintained and prevents usage due to old `psr/simple-cache`.
This removes the dependency on that package, which provided cache clearance on Classes and ParentClasses when their files changed.
The functionality can still be achieved by creating a custom implementation of `TheCodingMachine\GraphQLite\Utils\Cache\ClassBoundCacheContractFactoryInterface` that returns a wrapped version of `thecodingmachine/cache-utils` or an updated fork.

My implementation is quite naive and changes how the caching works. This may be considered a breaking change.

I'm happy to make adjustments if needed.

fixes thecodingmachine/graphqlite#693

Note: PHPStan seems to have problems. This is probably not related to my changes.

### DIFF
--- a/src/FactoryContext.php
+++ b/src/FactoryContext.php
@@ -6,6 +6,7 @@ namespace TheCodingMachine\GraphQLite;
 
 use Psr\Container\ContainerInterface;
 use Psr\SimpleCache\CacheInterface;
+use TheCodingMachine\GraphQLite\Cache\ClassBoundCacheContractFactoryInterface;
 use TheCodingMachine\GraphQLite\Mappers\RecursiveTypeMapperInterface;
 use TheCodingMachine\GraphQLite\Types\InputTypeValidatorInterface;
 use TheCodingMachine\GraphQLite\Types\TypeResolver;
@@ -31,6 +32,7 @@ final class FactoryContext
         private readonly InputTypeValidatorInterface|null $inputTypeValidator,
         private readonly int|null $globTTL,
         private readonly int|null $mapTTL = null,
+        private readonly ClassBoundCacheContractFactoryInterface|null $classBoundCacheContractFactory = null,
     ) {
     }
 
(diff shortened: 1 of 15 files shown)


src/FactoryContext.php
```diff
@@ -28,6 +29,7 @@ public function __construct(
         private readonly RecursiveTypeMapperInterface $recursiveTypeMapper,
         private readonly ContainerInterface $container,
         private readonly CacheInterface $cache,
+        private readonly ClassBoundCacheContractFactoryInterface|null $classBoundCacheContractFactory = null,
```
andrew-demb, line 32: Shoud it be added at the end of the list to preserve BC?

xyng: Good Catch! I added it near to cache for no particular reason (maybe optics?).
Will adjust later.

oojacoboo:
@xyng thanks for this.  For PHPStan, yea, not sure what's going on with that.  It looks like a syntax recommended change.  If you could give that a shot, that'd be great.

I think it'd be best to place this in `src/Cache` and not in the `src/Utils/Cache` directory.  `Utils` directories are like junk drawers, the mere existence of them just encourages more organization.

I think with @andrew-demb's recommendation on the `__constructor` ordering, there shouldn't be any BC issues.

xyng:
@oojacoboo @andrew-demb i have adjusted according to your feedback. Thanks a lot!
I also added a setter `setClassBoundCacheContractFactory` to the `SchemaFactory` to actually allow users to customize.

Issue #708, opened 2024-11-17
Is it intended BC break in unreleased version? Parameter annotations from method no more auto-bounded to parameters


andrew-demb:
In the `master` branch there's an unreleased (yet) commit that dropped support for Doctrine annotations - https://github.com/thecodingmachine/graphqlite/commit/870b4489f5bf5d789005c04e838a87dbbe8ea53f.

This commit also drops [1] a feature about allowing to use "method"-bound annotations (attributes) with explicit "target" (example: [2]).

The question is - is this dropping a feature intended? Are you open to reintroducing it - to simplify an upgrade path?

[1] https://github.com/thecodingmachine/graphqlite/commit/870b4489f5bf5d789005c04e838a87dbbe8ea53f#diff-b4cb271cb57cae8e3c9c55d3590bab4ea0781066f699d94ac970f89f8b7e06b4L356-L364
[2] https://github.com/thecodingmachine/graphqlite-symfony-validator-bridge/blob/2f677f6dc5c81660505e6c73952661d4d98f1199/tests/Fixtures/Controllers/UserController.php#L37

----

If I add this snippet locally, the logic about auto-bound annotations (attributes) from method to parameters will work as before.

```php
// \TheCodingMachine\GraphQLite\AnnotationReader::getParameterAnnotationsPerParameter
        $parameterAnnotationsPerParameter = [];

        $attributes = $method->getAttributes();
        foreach ($attributes as $attribute) {
            if (false === \is_a($attribute->getName(), ParameterAnnotationInterface::class, true)) {
                continue;
            }

            $parameterAnnotation = $attribute->newInstance();
            $parameterAnnotationsPerParameter[$parameterAnnotation->getTarget()][] = $parameterAnnotation;
        }
```

andrew-demb:
I think this is an unnecessary BC break for v7.

Here's a PR that reintroduces the feature https://github.com/thecodingmachine/graphqlite/pull/709

oojacoboo:
@andrew-demb dropping annotations wasn't a mistake and is the path forward.  Is there any reason why someone cannot switch to attributes?  There is even a rector that'll do it all for you...

https://getrector.com/blog/how-to-upgrade-annotations-to-attributes

[...]

PR #709, opened 2024-11-17
:sparkles: [#708] Re-introduce feature about supporting target method…

andrew-demb:
… attributes for resolving parameter annotations

Refs: #708 

Example use-case: https://github.com/thecodingmachine/graphqlite-symfony-validator-bridge/blob/2f677f6dc5c81660505e6c73952661d4d98f1199/tests/Fixtures/Controllers/UserController.php#L37

If you want to deprecate such behavior - please hint how to deprecate it properly (`symfony/deprecation-contracts`, `doctrine/deprecations`, just a comment in the code, etc)

### DIFF
--- a/src/AnnotationReader.php
+++ b/src/AnnotationReader.php
@@ -24,12 +24,14 @@ use TheCodingMachine\GraphQLite\Annotations\SourceFieldInterface;
 use TheCodingMachine\GraphQLite\Annotations\Type;
 use TheCodingMachine\GraphQLite\Annotations\TypeInterface;
 
+use function array_diff_key;
 use function array_filter;
 use function array_key_exists;
 use function array_map;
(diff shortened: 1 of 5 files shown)


andrew-demb:
@oojacoboo, please let me know if I need to change anything here

oojacoboo:
@andrew-demb please use the `@deprecated` tag - makes it easily searchable.

andrew-demb:
@oojacoboo but I need to deprecate only using "method target" annotation to be used for parameters.

I don't understand how to deprecate one feature from changed method logic, not  all the method.

I will add "trigger_error" with E_USER_DEPRECATED just like in other places in graphqlite

oojacoboo:
@andrew-demb That works - yep

Issue #711, opened 2024-11-18
Problem with reusing the same cache entries for differently configured class finders


andrew-demb:
I'm in progress of making Symfony bundle `thecodingmachine/graphqlite-bundle` compatible with `thecodingmachine/graphqlite` v7.

While upgrading the Symfony bundle codebase, phpunit tests showed a problem with functionality that relies upon adding a custom type mapper factory (`\TheCodingMachine\GraphQLite\Mappers\StaticClassListTypeMapperFactory`) to the main Schema in `dev-master` version of `graphqlite` (`dev-master` has commits that improved performance and caching of the library https://github.com/thecodingmachine/graphqlite/pull/698). The problem was that `graphqlite` failed to find the GraphQL type, that should have been added by custom type mapper (instantiated via factory).

After investigating, I found the place, where `graphqlite` while using cache for class finders (`\TheCodingMachine\GraphQLite\Discovery\ClassFinder`) - doesn't try to change the cache key for different instances of a class finder.

https://github.com/thecodingmachine/graphqlite/blob/d45b9b9555f360d995ebf3cb38d0cf7999bd394e/src/Mappers/ClassFinderTypeMapper.php#L59

https://github.com/thecodingmachine/graphqlite/blob/d45b9b9555f360d995ebf3cb38d0cf7999bd394e/src/Discovery/Cache/SnapshotClassFinderComputedCache.php#L52

https://github.com/thecodingmachine/graphqlite/blob/d45b9b9555f360d995ebf3cb38d0cf7999bd394e/src/Discovery/Cache/HardClassFinderComputedCache.php#L35

I will create a PR that fixes the behavior (but won't reuse the cache for differently configured instances of class finder) with tests.

/cc @oprypkhantc @SCIF

PS: WIP PR for `graphqlite-bundle` on my fork: https://github.com/andrew-demb/graphqlite-bundle/pull/1

PR #712, opened 2024-11-18
:bug: [#711] Don't reuse cache for differently configured class finder instances

andrew-demb:
Context: #711 

/cc @oprypkhantc @SCIF


oprypkhantc (approved):

oprypkhantc:
Makes sense; I thought to add something of the kind, but didn't think there was a real use case for that.

PR #721, opened 2024-12-18
Fixed Adapter class for compatibility with base ObjectType

### LINKED ISSUE #697: Should MutableObjectTypeAdapter be setting description?
I'm just trying to get my head around this library works, I tried adding a static type defition mapping like so:

```
$myStaticTypeMapper = new StaticTypeMapper();
$myStaticTypeMapper->setTypes([
	Example::class => new ExampleOutputType(),
]);
$schemaFactory->addTypeMapper($myStaticTypeMapper);
```

When I run 
`SchemaPrinter::doPrint($schema);`

And I get an error that there is no `description` set on the `MutableObjectTypeAdapter` that is wrapping `ExampleOutputType`.  It appears the printer calls $def->description even though the mutable object type adapter never sets a description.

Guessing i'm trying to use the static type mapper incorrectly, but figured i'd ask, thanks!

oojacoboo:
Fixes #697

### DIFF
--- a/src/Mappers/Proxys/MutableAdapterTrait.php
+++ b/src/Mappers/Proxys/MutableAdapterTrait.php
@@ -39,13 +39,13 @@ trait MutableAdapterTrait
         $this->type->assertValid();
     }
 
-    public function jsonSerialize():string
+    public function jsonSerialize(): string
     {
         return $this->type->jsonSerialize();
     }
 
 
-    public function toString():string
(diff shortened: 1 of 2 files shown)


oojacoboo:
@cvergne thanks for catching that.  I originally started the release with 8.0.1, but decided to update for 8.1.0 b/c of PHP 8.4 support.  Tag fixed now.

PR #739, opened 2025-03-15
Allow callable() as field deferring mechanism

### LINKED ISSUE #738: Cleaner and simpler field deferring
Hey

We've found use cases for deferring field resolution directly, without parameters or prefetch, e.g.:

```php
#[Type]
class SomeType {
    #[Field(outputType: 'Int!')]
    public function complexCalculation(
        bool $arg1
    ): Deferred {
        // Do something with $arg1, similar to prefetch

        // Return value from some kind of buffer, again, similar to prefetch
        return new Deferred(fn () => 123);
    }
}
```

Currently it's possible, but ugly: 
- requires the use of `outputType`
- forces the return type to be `GraphQL\Deferred`
- forces creation of `new GraphQL\Deferred` instance everywhere

I propose that such cases are simplified into returning a (typed) callable, which removes the dependency on `Deferred` from userland code, as well as allows specifying an actual PHPDoc type:

```php
#[Type]
class SomeType {
    /** @return callable(): int */
    #[Field]
    public function complexCalculation(
        bool $arg1
    ): callable {
        // Do something with $arg1, similar to prefetch

        // Return value from some kind of buffer, again, similar to prefetch
        return fn () => 123;
    }
}
```

The implementation should be somewhat trivial - under the hood the callable value would just get wrapped in `Deferred`, and additional root type mapper would handle `callable()` PHPDoc types. Thoughts?

oprypkhantc:
Closes #738


src/Mappers/CannotMapTypeException.php
```diff
@@ -177,4 +180,19 @@ public static function createForNonNullReturnByTypeMapper(): self
     {
         return new self('a type mapper returned a GraphQL\Type\Definition\NonNull instance. All instances returned by type mappers should be nullable. It is the role of the NullableTypeMapperAdapter class to make a GraphQL type in a "NonNull". Note: this is an error in the TypeMapper code or in GraphQLite itself. Please check your custom type mappers or open an issue on GitHub if you don\'t have any custom type mapper.');
     }
+
+    public static function createForUnexpectedCallableParameters(): self
```
oprypkhantc, line 184: Not a fan of having these (type mapper specific) constructors here, but that's the way it's done for every other mapper 🤔

composer.json
```diff
@@ -13,8 +13,8 @@
         "php": ">=8.1",
         "ext-json": "*",
         "composer/package-versions-deprecated": "^1.8",
-        "phpdocumentor/reflection-docblock": "^4.3 || ^5.0",
-        "phpdocumentor/type-resolver": "^1.4",
+        "phpdocumentor/reflection-docblock": "^5.4",
```
oprypkhantc, line 16: 4.3 was not allowed either way; even with `--prefer-lowest` it would use 5.0 or 5.1. So not a breaking change

tests/Fixtures/Integration/Models/Blog.php
```diff
@@ -36,13 +36,17 @@ public function getPosts(
...
+    /**
+     * @param Blog[][] $prefetchedSubBlogs
+     *
+     * @return callable(): Blog[]
+     */
+    #[Field]
     public function getSubBlogs(
         #[Prefetch('prefetchSubBlogs')]
         array $prefetchedSubBlogs,
-    ): Deferred {
```
oojacoboo, line 44: Let's keep this test as well, instead of replacing it.

oojacoboo:
@oprypkhantc so I've merged this, but am getting an error locally.

>current() expects exactly 1 argument, 0 given
`vendor/webonyx/graphql-php/src/Executor/Promise/Adapter/SyncPromise.php:63`

Any ideas on this?

oojacoboo:
@oprypkhantc @andrew-demb 

```gql
book(id: "%s") {
    id
    authors {
        addresses {
            type {
                key
            }
        }
    }
}
```

This call in the stack looks particularly interesting:

```
{
      "file": "vendor/webonyx/graphql-php/src/Server/Helper.php",
      "line": 204,
      "function": "wait",
      "class": "GraphQL\\Executor\\Promise\\Adapter\\SyncPromiseAdapter",
      "type": "->",
      "args": [
        {
          "adoptedPromise": {
            "state": "fulfilled",
            "result": {
              "errors": [
                {
                  "message": "Internal server error",
                  "locations": [{ "line": 8, "column": 33 }],
                  "path": ["book", "authors", 0, "addresses", 0, "type", "key"]
                }
              ]
            }
          }
        }
      ]
},
```

So, on line 63 of `\GraphQL\Executor\Promise\Adapter\SyncPromise`, it's trying to resolve the promise `current()`.  When debugging the exception there.  That's the value of `$executor`.

andrew-demb:
@oojacoboo can you share code of the "address type" graphql type?

Is it return any callable in some field?

oojacoboo:
@andrew-demb sure - quite simple class

```php
#[GraphQLite\Type]
#[GraphQLite\Input]

class AddressType
{

    public const string BILLING = 'billing';
    public const string CURRENT = 'current';
    public const string LEGAL = 'legal';
    public const string MAILING = 'mailing';
    public const string PHYSICAL = 'physical';


    #[GraphQLite\Field(for: 'AddressTypeInput')]
    protected string $key;


    public function __construct(string $key)
    {
        $this->key = $key;
    }


    public function __toString(): string
    {
        return $this->getKey();
    }


    #[GraphQLite\Field(outputType: 'ID!')]
    public function getKey(): string
    {
        return $this->key;
    }
}
```

oprypkhantc:
@oojacoboo Could it be that you updated other dependencies alongside `thecodingmachine/graphqlite` which caused the issue?

autaut03:
I found one issue by pure luck. It might be related to your error. I'll send a PR today

Issue #729, opened 2025-02-14
Prefetching nested fields with shared Type instances throws error


cdomigan:
Thank you for this wonderful library.

Until the change introduced by #702 by @sudevva I was able to "share" instances of Types in my prefetch logic.

The change to this line: https://github.com/thecodingmachine/graphqlite/blob/37c06a9ee307980fd6be0f423d789412eda7c9f6/src/Parameters/PrefetchDataParameter.php#L62

means that the source is now purged from results. If I then want to return the same Type instance later, I will get an error:
```
UnexpectedValueException: Object not found
/workspaces/graphqlite/src/PrefetchBuffer.php:90
```

This error only shows up in a nested prefetch context, specifically *where an intermediate type is being prefetched, instances of that type are being shared, and it has a prefetched child field also*.

In the test suite on this repo, this demonstrates the issue:

```php
// tests/Integration/EndToEndTest.php

// A NEW TEST TO DEMONSTRATE THE ISSUE
    public function testPrefetchingDeeplyNested(): void
    {
        $schema = $this->mainContainer->get(Schema::class);
        assert($schema instanceof Schema);

        $schema->assertValid();

        $queryString = '
        query {
            companies {
                contact {
                    name
                    supervisor {
                        name
                    }
                }
            }
        }
        ';

        $result = GraphQL::executeQuery(
            $schema,
            $queryString,
            null,
            new Context(),
        );

        $this->assertSame([
            'companies' => [
                [
                    'contact' => [
                        'name' => 'Kate',
                        'supervisor' => [
                            'name' => 'The Boss',
                        ],
                    ],
                ],
                [
                    'contact' => [
                        'name' => 'Kate',
                        'supervisor' => [
                            'name' => 'The Boss',
                        ],
                    ],
                ],
            ]
        ], $this->getSuccessResult($result));
    }
```

```php
// tests/Fixtures/Integration/Controllers/CompanyController
/**
     * @return Company[]
     */
    #[Query]
    public function getCompanies(): array
    {
        return [new Company('Company1'), new Company('Company2')];
    }
```

```php
// tests/Fixtures/Integration/Types/CompanyType.php
    /**
     * @param Company[] $companies
     *
     * @return Contact[]
     */
    public static function prefetchContacts(array $companies): array
    {
        $contacts = [];

        $kate = new Contact('Kate');

        foreach ($companies as $company) {
            // This works
            //$contacts[$company->name] = new Contact('Kate');

            // This does not work
[...]

PR #752, opened 2025-07-30
don't purge the results from buffer

januszmk:
I tried upgrading graphqlite lib in my project and I was hit by this issue: https://github.com/thecodingmachine/graphqlite/issues/729

I am making pr that removes purging from the buffer, but if there is different way to fix the issue, please let me know


oojacoboo:
@oprypkhantc was there a reason to purge results from the prefetch buffer?  I'm inclined to merge this, but would like to hear your reasoning for adding.

autaut03:
We should get a new context (along with the whole prefetch buffer) on each new request, so it will get cleared, just a bit later. It should result in higher memory usage before the response is generated, right?

PR #753, opened 2025-07-31
Fix callable type mapping to be Closure mapping

oprypkhantc:
Quite accidentally I discovered a bug with my implementation: if a value that is also a callable is returned from a field, it will get wrapped in a `Deferred` wrapper and be called as a callable, instead of simply being returned as a value.

E.g.:

```php
function test() {
    throw new RuntimeException('This is called!');
}

#[Type]
class Test {
    #[Field]
    public string $title = 'test';
}

class TestController {
    #[Query]
    public function test(): Test {
        return new Test();
    }
}
```

If you then run the `test { title }` query, instead of simply returning the `{"title": "test"}`, it would call the `test` function - because `is_callable('test')` returns `true`.

## Potential fixes

This could be solved two ways: 
1. remove support for `callable`, only leave support for actual `Closure`
2. keep support for `callable`, but make it so it wraps values in `Deferred` based on whether the type was specified as callable, not if the value itself is callable

This PR implements option #1. Option #2 would require some architectural changes to how GraphQLite works, way outside the scope of this feature :)


oojacoboo (approved):

oojacoboo:
@oprypkhantc tagged!

PR #766, opened 2025-10-31
Add hasser as property accessor

adynemo:
Hi! :wave: 

This PR allows using hasser as property accessor. Hasser is another type of "getter" and acts like an isser.

Considering this following class:
```php
class MyClass
{
  private bool $something;

  public function hasSomething(): bool
  {
    return $this->something;
  }
}
```

Now, we have to override the getter:
```php
#[Type(class: MyClass::class)]
class MyClassType
{
    #[Field]
    public function getSomething(MyClass $myClass): bool
    {
        return $myClass->hasSomething();
    }
}
```

With that PR:
```php
#[Type(class: MyClass::class)]
#[SourceField(name: 'something')]
class MyClassType
{
}
```

Just for information, I've made a search on the issue tracker and didn't find an issue about hassers and also run PHPStan and PHPUnit.

Thanks :slightly_smiling_face:

### DIFF
--- a/src/Utils/PropertyAccessor.php
+++ b/src/Utils/PropertyAccessor.php
@@ -22,7 +22,7 @@ class PropertyAccessor
      */
     public static function findGetter(string $class, string $propertyName): string|null
     {
-        foreach (['get', 'is'] as $prefix) {
+        foreach (['get', 'is', 'has'] as $prefix) {
             $methodName = self::propertyToMethodName($prefix, $propertyName);
 
             if (self::isPublicMethod($class, $methodName)) {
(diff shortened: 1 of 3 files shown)


oojacoboo:
@adynemo after some additional consideration on this, I think we can implement it, but with the addition of a new SchemaFactory config, like:

```php
$factory = new SchemaFactory(...);
$factory->stripFieldPrefixes(['is', 'has', 'get', 'set']); // maybe a second param could be offered for ensuring the next character in the field name (that isn't stripped) is always lowercase, defaulting to true.
```

You'll need to rebase this PR branch and we'll need sufficient test coverage for this.  As for the default, I guess we should keep it how it is now, to avoid BC breaks.  If you have any feedback on that, happy to discuss.

oojacoboo:
Superseded by #827

PR #779, opened 2026-01-13
fix: allow adding properties starting with 'set'

supresys:
Currently, valid properties with the 'set' prefix are removed from the schema: **set**tled, **set**tings, **set**up, **set**OfParameters, etc. Also, all setters are cut out despite the presence of an explicit 'Field' indication. This is clearly incorrect behavior.

This PR shouldn't break backwards compatibility because it requires explicitly specifying 'Field' attribute.


oojacoboo:
Thanks.  So, why not just check to see if the `name` param has been defined on the `Field` attribute, explicitly, and use it if so, instead of checking to see if the method name starts with `set`?

supresys:
> So, why not just check to see if the `name` param has been defined on the `Field` attribute, explicitly, and use it if so, instead of checking to see if the method name starts with `set`?

I didn't quite understand the question. I'm not adding this check, I'm removing it! The question should be addressed to whoever added this check to the code that's causing the fields to disappear from the schema. I completely agree that the name should be taken from the 'name' property. My PR enables this behaviour.

supresys:
I've added a test that covers this issue and its solution.

PR #789, opened 2026-04-07
Make sure that types are frozen before creating the schema

michael-georgiadis:
# Description

Some interface types escape from the type freezing. That leads to the type being added in the schema, but whenever you try to validate the schema validation/generation tries to fetch the fields of an unfrozen type which leads to an exception of the `MutableTrait`. 

This is referenced in [the issue 308](https://github.com/thecodingmachine/graphqlite/issues/308)

### DIFF
--- a/src/SchemaFactory.php
+++ b/src/SchemaFactory.php
@@ -538,6 +538,8 @@ class SchemaFactory
 
         $aggregateQueryProvider = new AggregateQueryProvider($queryProviders);
 
+        $typeRegistry->finalizeTypes();
+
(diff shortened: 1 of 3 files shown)


oojacoboo:
@michael-georgiadis Thanks for this!  Have a look at https://github.com/thecodingmachine/graphqlite/pull/790 I was working on at the same time as you, apparently.  I saw your comment earlier that you'd deleted (Github sent it), and it got me thinking.  Let me know your thoughts.

michael-georgiadis:
Thanks man, closing this one then!

PR #792, opened 2026-04-18
Fix #[Security] on ExtendType fields (source-injection array_combine crash)

oojacoboo:
## Problem

`#[Security]` on an `#[ExtendType]` field method throws at runtime with:

```
TypeError: array_combine(): Argument #1 (\$keys) and argument #2 (\$values)
must have the same number of elements
```

In most setups this masks as a generic `"Internal server error"` from any `#[Security]`-decorated field on an `#[ExtendType]`, which isn't discovered until someone tries to guard an ExtendType field by role.

## Root cause

`SecurityFieldMiddleware::process()` captures `\$parameters` *before* `QueryField::fromFieldDescriptor` prepends a `SourceParameter` (it does this when `isInjectSource()` is true). At resolver invocation time `\$args` therefore includes the source as its first element while `\$parameters` (captured earlier) does not, so the `array_combine(\$argsName, \$args)` call fails.

`AuthorizationFieldMiddleware` (`@Logged` / `@Right`) sidesteps the issue by using `function (...\$args) use (...)` and passing args through transparently — it doesn't need to map args to parameter names because it has no expression language context. The Security middlewares are the only ones that zip args and parameters together, so they're the only ones that need the fix.

The same bug exists in `SecurityInputFieldMiddleware` for input field factories where source is similarly injected by `InputField::fromFieldDescriptor`.

## Fix

Pass `isInjectSource()` through to `getVariables()` and slice the leading source arg off \`\$args\` before the `array_combine`. The source is still available via `this` in the expression context, so no information is lost for Security expressions.

## Tests

New regression test in \`EndToEndTest::testEndToEndSecurityOnExtendTypeField\` — adds a \`#[Security]\`-decorated field to the existing \`ExtendedContactType\` fixture and verifies both the \`failWith: null\` path (no user logged in) and the authorized path (user.bar == 42) resolve correctly. Before the fix the test throws the \`array_combine\` \`TypeError\`; after, it passes.

All existing tests continue to pass (494/494).

## Checklist

- [x] Reproduces with existing fixtures/schema
- [x] Fix applied to both \`SecurityFieldMiddleware\` and \`SecurityInputFieldMiddleware\` (same bug pattern)
- [x] Source still available via \`this\` in Security expressions (no regression for existing users)
- [x] Regression test added
- [x] Full test suite passes

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         install-args: ['', '--prefer-lowest']
-        php-version: ['8.1', '8.2', '8.3', '8.4']
+        php-version: ['8.2', '8.3', '8.4']
       fail-fast: false
 
     steps:
@@ -73,7 +73,7 @@ jobs:
         run: "composer phpstan"
 
       - name: "Run coding standard checks with squizlabs/php_codesniffer on minimum supported PHP version"
-        if: matrix.php-version == '8.1'
+        if: matrix.php-version == '8.2'
         run: composer cs-check
 
       - name: "Archive code coverage results"
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         install-args: ['', '--prefer-lowest']
-        php-version: ['8.2', '8.3', '8.4']
+        php-version: ['8.2', '8.3', '8.4', '8.5']
       fail-fast: false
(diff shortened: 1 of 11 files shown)


oojacoboo:
This PR removes support for PHP 8.1, following official EOL security support.  Additionally, PHP 8.5 is now officially supported and added to the test matrix.

charly-ffss:
Hi @oojacoboo, when do you think it can be released ?

oojacoboo:
@charly-ffss done

charly-ffss:
Thank you very much sir 😄

PR #801, opened 2026-04-24
Fix Security "this" variable on ExtendType fields

### LINKED ISSUE #653: Security attribute on extended type
I am getting error when trying to use Security attribute on extended type class.

I have mutation class
```
final class RealtyMuation {
    // some methods
}
```

Then I have extended class, with Security attribute usage
```
#[ExtendType(RealtyMutation::class)]
final class RealtyCheckupMutation {
    #[Field]
    #[Security(
        expression: "is_granted('CAN_EDIT_REALTY', realtyId)",
        failWith: new AccessDeniedException(),
    )]
    public funtion startTechnicalException(RealtyMutation $mutation, UuidInterface $realtyId) 
    {
    }
}

Which gives me error
```
array_combine(): Argument #1 ($keys) and argument #2 ($values) must have the same number of elements
```
on line https://github.com/thecodingmachine/graphqlite/blob/b63eaec119e6bf01146e5ca74795b79d98fbc97e/src/Middlewares/SecurityFieldMiddleware.php#L118


Version: v5.0.3

charly-ffss:
Fixes #653
Related to #792

## Problem

When using `#[Security]` on a field declared in an `#[ExtendType]`, the `this`
variable incorrectly points to the resolver instance instead of the source object.

This breaks authorization logic relying on the domain object (e.g. voters).

## Fix

Preserve `$source` when `injectSource=true` instead of always using
`$originalResolver->executionSource($source)`.

## Test coverage

Extends existing end-to-end test to assert that:

- `this` is the source object
- not the resolver instance

## Result

- Correct behavior for `#[Security]` on `#[ExtendType]`
- No regression on existing tests

### DIFF
--- a/src/Middlewares/SecurityFieldMiddleware.php
+++ b/src/Middlewares/SecurityFieldMiddleware.php
@@ -83,7 +83,12 @@ class SecurityFieldMiddleware implements FieldMiddlewareInterface
         $injectSource = $queryFieldDescriptor->isInjectSource();
 
         $queryFieldDescriptor = $queryFieldDescriptor->withResolver(function (object|null $source, ...$args) use ($originalResolver, $securityAnnotations, $resolver, $failWith, $parameters, $queryFieldDescriptor, $injectSource) {
-            $variables = $this->getVariables($args, $parameters, $originalResolver->executionSource($source), $injectSource);
+            $variables = $this->getVariables(
+                $args,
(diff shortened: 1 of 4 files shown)


oojacoboo (approved):

oojacoboo:
Thanks @charly-ffss!

PR #807, opened 2026-06-02
feat: Add directive layer

michael-georgiadis:
# Description

As per your comment on [this PR](https://github.com/thecodingmachine/graphqlite/pull/806), I am splitting the implementation into two. 

I am focusing on configuring the directive layer on this one. Specifying two directives to start with.

## How it works

A directive is a PHP attribute implementing one of the family interfaces, which ties it to a GraphQL location (`FieldDirective` → `FIELD_DEFINITION`, `InputObjectTypeDirective` → `INPUT_OBJECT`, etc.). Implementing the marker interface alone is metadata-only;  implementing the matching `Behavioral*` sub-interface adds an apply hook that runs through a middleware pipe during type generation.

## The built-ins on the layer
                                                                                                                                                                                                                                                                                                                   
- **`#[OneOf]`** → sets webonyx's `isOneOf` flag → prints `input LookupInput @oneOf`.
- **`#[Deprecated(reason:)]`** → sets the field's deprecation reason → prints
`legacy: String! @deprecated(reason: "...")`.

Both bind to directives webonyx already declares, so they print through webonyx's own SDL output and we don't register duplicate definitions for them (`DirectiveDefinition::$builtIn`).

### `@deprecated`: additive, not a rewrite

The existing docblock `@deprecated` support is untouched. `#[Deprecated]` is an attribute-based
path that composes with it rather than fighting it:

| Declaration | Result |
|---|---|
| `#[Deprecated(reason: 'X')]` (± docblock) | `@deprecated(reason: "X")` — explicit wins |
| `#[Deprecated]` + `/** @deprecated Y */` | `@deprecated(reason: "Y")` — docblock kept |
| `#[Deprecated]`, no docblock | `@deprecated(reason: "No longer supported")` — default |

## Usage validation

PHP's `#[Attribute]` targets can't tell a `#[Type]` class from an `#[Input]` class (both are `TARGET_CLASS`), so a misplaced `#[OneOf]` on a `#[Type]` would otherwise be silently dropped by the interface-based collection. `DirectiveValidator` catches that and throws a clear `InvalidDirectiveException` instead.


michael-georgiadis:
Hey @oojacoboo, I don't want this to become stale. Is it something that you need to get it over the finish line?

oojacoboo:
Well, for starters, the tests are all going to have to pass.  Beyond that, an hour of my time to review it.

andrew-demb:
> Well, for starters, the tests are all going to have to pass.

see https://github.com/thecodingmachine/graphqlite/pull/811

oojacoboo:
@michael-georgiadis #811 was merged.  Please rebase and hopefully that resolves tests.

michael-georgiadis:
@oojacoboo Done ✅ ! 

I think the most meaningful cases are covered and I also included an end-to-end one. However, if you want more coverage on specific sub-units happy to add more 🫡

michael-georgiadis:
@oojacoboo Any updates on the review? 👀

src/SchemaFactory.php
```diff
@@ -558,7 +579,14 @@ classBoundCache: $classBoundCache,
...
         $aggregateQueryProvider = new AggregateQueryProvider($queryProviders);
 
-        return new Schema($aggregateQueryProvider, $recursiveTypeMapper, $typeResolver, $topRootTypeMapper, $this->schemaConfig);
+        return new Schema(
+            $aggregateQueryProvider,
+            $recursiveTypeMapper,
+            $typeResolver,
+            $topRootTypeMapper,
+            $this->schemaConfig,
+            $directiveRegistry->webonyxDirectives(),
```
oojacoboo, line 588: I assume this will be updated when we support custom directives?

oojacoboo (changes requested):
@michael-georgiadis this looks really good - thanks!  The tests need some additional work though.  Firstly, there are custom directives that are still lingering here.  Also, there aren't any E2E tests and that's a requirement to get this merged.

After you get that resolved, I think we can get this merged.

src/SchemaFactory.php
michael-georgiadis: Exactly. Then we would have to fetch all the built-ins and also the custom directives specified by the user

michael-georgiadis:
@oojacoboo There was already an E2E test (`DirectiveLayerEndToEndTest.php`). But you were right in the fact that I was not testing if it actually was applying the behaviors when the users executes queries. So I enriched it to also cover that. 

Other than that, I also removed the tests that were custom directive related

oojacoboo (changes requested):
@michael-georgiadis thanks.  I still need E2E GQL tests that validate the response.

michael-georgiadis:
Hey @oojacoboo, I added some that validate the response. Tell me what you think

michael-georgiadis:
@oojacoboo any updates on this?

oojacoboo:
@michael-georgiadis we need E2E GQL tests for all directives, built-in and custom, covering all the various scenarios of use.  Deprecation looks fine, but we need a couple OneOf tests.

michael-georgiadis:
@oojacoboo added the oneOf tests. Also did a fix that the e2e surfaced. There are no custom directives yet. That will come in a later PR

oojacoboo:
@michael-georgiadis thanks.  Would you mind resolving those other errors so tests pass?  I realize it's outside the scope of this PR.  If you're unsure, I can get another PR in place to resolve those soon.

michael-georgiadis:
@oojacoboo identified the issue and is now fixed. It was indeed unrelated to the changes. Dropped the named argument in favor of positional approach on the `ArrayAdapter` initialization. 

If you want to make it more explicit, we can use the `version_compare` function, but I don't think it's worth it for this case

oojacoboo (approved):

oojacoboo:
@michael-georgiadis thanks for working through this PR with me.  We can work on getting the custom directives merged now.

PR #817, opened 2026-07-15
fix: support #[Type] interfaces used as output types

### LINKED ISSUE #816: `#[Type]` interface used as an output type breaks schema build on 8.3.0 (worked on 8.1.2)
# Description

Ran into what looks like a regression somewhere between 8.1.2 and 8.3.0.

Short version: declare a GraphQL interface the documented way (#[Type] on a PHP interface), have something return that interface as an output type, and building the schema blows up on 8.3.0. The exact same code builds fine on 8.1.2.

Error:                                                                                                                                                                             
Expected GraphQL type "Foo" to be an MutableObjectType. Got a TheCodingMachine\GraphQLite\Types\TypeAnnotatedInterfaceType thrown from TypeRegistry::getMutableObjectType().

Repro is basically the interfaces doc example:
```php
  #[Type(name: 'Foo')]                                                                                                                                                               
  interface FooInterface { #[Field] public function getId(): ID; }                                                                                                                   
                                                                                                                                                                                     
  #[Type]                                                                                                                                                                            
  class Bar implements FooInterface {                                                                                                                                                
      public function __construct(private Thing $thing) {}                                                                                                                           
      public function getId(): ID { return new ID((string) $this->thing->id); }                                                                                                      
  }
```

Then have any field return Foo (the interface) as its output type and build/validate the schema.
What's going on, from a backtrace:
```
Schema->assertValid()                                                                                                                                                              
 -> RecursiveTypeMapper::getOutputTypes()        (RecursiveTypeMapper.php:447)
   -> mapClassToType(FooInterface::class)        // object-only
     -> TypeRegistry::getMutableObjectType('Foo')   (TypeRegistry.php:80)
     -> it's an interface -> throws
```
`getOutputTypes()`loops over every supported class and calls mapClassToType() on it, interfaces included. In 8.3.0 mapClassToType() is typed `: MutableObjectType`, so the second it hits the interface class it throws. On 8.1.2 that method could return an interface, so it was fine.

michael-georgiadis:
Closes #816 

# Description

`#[Type]` on a PHP interface is documented as a supported way to declare a GraphQL interface, but if that interface is used as an output type, building the schema blows up:

```
Expected GraphQL type "X" to be an MutableObjectType. 
Got a TheCodingMachine\GraphQLite\Types\TypeAnnotatedInterfaceType
```

It only shows up once the interface has already been registered (e.g. it was referenced by name via an `outputType` before the full type map is built), so it tends to surface on `$schema->assertValid()` rather than on a narrow query.

This was happening, because `getOutputTypes()` calls the object-only `mapClassToType()` on every supported class, interfaces included. Its "already registered" branch then calls `getMutableObjectType()`, which throws on an interface.                                                                                                             

## Fix
When the mapped type is a `MutableInterfaceType`, return its generated concrete companion instead (same conversion the method already does on the non-cached path).

### DIFF
--- a/src/Mappers/RecursiveTypeMapper.php
+++ b/src/Mappers/RecursiveTypeMapper.php
@@ -308,15 +308,19 @@ class RecursiveTypeMapper implements RecursiveTypeMapperInterface
 
         // In the event this type was already part of cache, let's not extend it.
         if ($this->typeRegistry->hasType($type->name)) {
+            // Already-registered interface: return its concrete companion (as the non-cached path
+            // does below), else getMutableObjectType() throws on it.
+            if ($type instanceof MutableInterfaceType) {
+                return $this->classToTypeCache[$cacheKey]
+                    ??= $this->getGeneratedObjectTypeFromInterfaceType($type);
+            }
+
             $cachedType = $this->typeRegistry->getMutableObjectType($type->name);
             if ($cachedType !== $type) {
                 throw new RuntimeException('Cached type in registry is not the type returned by type mapper.');
             }
 
-            //if ($cachedType->getStatus() === MutableObjectType::STATUS_FROZEN) {
-                return $type;
-
-            //}
+            return $type;
         }
 
         $this->typeRegistry->registerType($type);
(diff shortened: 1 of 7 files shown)


oojacoboo:
@michael-georgiadis thanks for this!

oojacoboo (approved):

PR #819, opened 2026-07-22
fix: Undefined array input type

michael-georgiadis:
# Description 

  Optional-via-`Undefined` works for scalars but explodes on lists:
                                                                                                                                                                                                                                                                                                                            
  ```php                                                                                                                                                                                                                                                                                                                    
  /** @param list<string>|Undefined|null $ids */
  public readonly array|Undefined|null $ids = Undefined::VALUE,
  // CannotMapTypeException: type-hinted to "array|list<string>" ... forbidden
```
  
Fix: drop Undefined and take the non-null core before the container check, reusing existing helpers (UndefinedTypeMapper + a now-public static getNonNullableType).

Covers array/list. Native iterable isn't handled (PHP expands it to array|Traversable) and since it's rare for an input, I left it alone

### DIFF
--- a/tests/Integration/EndToEndTest.php
+++ b/tests/Integration/EndToEndTest.php
@@ -2015,6 +2015,34 @@ class EndToEndTest extends IntegrationTestCase
         $this->assertSame('test', $data['updateArticle']['localizedTitle']);
     }
 
+    public function testEndToEndInputUndefinedListValue(): void
+    {
+        $schema = $this->mainContainer->get(Schema::class);
+        assert($schema instanceof Schema);
+
+        // A provided list round-trips.
+        $result = GraphQL::executeQuery($schema, '
+        mutation {
+            updateArticle(input: { tags: ["news", "tech"] }) {
+                tags
+            }
+        }
+        ');
+        $data = $this->getSuccessResult($result);
+        $this->assertSame(['news', 'tech'], $data['updateArticle']['tags']);
+
+        // An omitted list is Undefined, so the controller leaves it untouched.
+        $result = GraphQL::executeQuery($schema, '
+        mutation {
+            updateArticle(input: { magazine: "The Verge" }) {
+                tags
+            }
+        }
+        ');
+        $data = $this->getSuccessResult($result);
+        $this->assertNull($data['updateArticle']['tags']);
+    }
+
     public function testEndToEndSchemaIsPrintable(): void
     {
         $this->expectNotToPerformAssertions();
--- a/tests/Integration/EndToEndTest.php
+++ b/tests/Integration/EndToEndTest.php
@@ -2031,10 +2031,21 @@ class EndToEndTest extends IntegrationTestCase
         $data = $this->getSuccessResult($result);
         $this->assertSame(['news', 'tech'], $data['updateArticle']['tags']);
 
-        // An omitted list is Undefined, so the controller leaves it untouched.
+        // An omitted list is Undefined, so the controller leaves the default untouched.
         $result = GraphQL::executeQuery($schema, '
         mutation {
             updateArticle(input: { magazine: "The Verge" }) {
+                tags
+            }
+        }
+        ');
+        $data = $this->getSuccessResult($result);
+        $this->assertSame(['tech', 'news'], $data['updateArticle']['tags']);
+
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by oojacoboo** [comment]
@michael-georgiadis I'm not quite sure I understand the issue here.  Where are things breaking?

**Comment by michael-georgiadis** [comment]
@oojacoboo So, up until now the library handled the undefined compounds on the non-doc-block based type hints. 

An `int|Undefined|null` for example would work fine. 

But for this case
```php
/** @param list<int> $array */
private array|Undefined|null $array;
```

if the code block below does not exist:
```php
  if ($innerType instanceof Compound && UndefinedTypeMapper::containsUndefined($innerType)) {                                                                                                                                                                                                                               
      $innerType = NullableTypeMapperAdapter::getNonNullableType(                                                                                                                                                                                                                                                           
          UndefinedTypeMapper::replaceUndefinedWith($innerType),                                                                                                                                                                                                                                                            
      ) ?? $innerType;                                                                                                                                                                                                                                                                                                      
  }   
```

The type coming in in this if statement
```php
        if (
            $innerType instanceof Array_
            || $innerType instanceof Iterable_
            || $innerType instanceof Mixed_
            || $innerType instanceof Callable_
            || ($innerType instanceof Object_
                && $docBlockType instanceof Collection
                && (string) $innerType->getFqsen() === (string) $docBlockType->getFqsen()
            )
        )
```
was a `Compound(array, Undefined, null)`. That would make it go to the else statement and blow up during schema generation. 

So the fix here is basically unwrapping the compound with previously written code and making sure that the innerType is handled as an Array instead and prints the type correctly. 

Scalars also work by passing the else statement, but there are no conflicting docblocks so they survive

michael-georgiadis:
This is not me being passive aggressive btw, but you can checkout the branch and try the integration tests with the fixtures but reverting the actual code-change and you will see that it blows up

oojacoboo:
So, I guess I'm uncertain why you'd want to docblock type differently from the actual type signature.

michael-georgiadis:
@oojacoboo I am unsure what you mean by that. I will try my best to clear the uncertainty tho. The docblock and the actualy type signature for arrays carry different information (as ofc you know). 

So, the only thing that `array` gives you is basically that it's a list. But for GQLite to know which scalar or type is in that list you'd need the docblock type hint.

oojacoboo:
```php
/** @param list<int> $array */
private array|Undefined|null $array;
```

I'm saying this is an incorrect phpdoc type, as it doesn't include `Undefined` or `null`.  Instead, it should be:

```php
/** @param list<int>|Undefined|null $array */
private array|Undefined|null $array;
```

michael-georgiadis:
@oojacoboo fair point. Tried it with the proper docblock typing (as you can see in the latest commit on the fixtures) and the issue still remains if I revert my code.

tests/Integration/EndToEndTest.php
oojacoboo, line 1: Can you add a test here that checks that a passed `null` value into the input type is respected as an update?  So, the default value would be something other than `null`, and it's updated to be nullified.

oojacoboo (changes requested):

tests/Integration/EndToEndTest.php
michael-georgiadis: @oojacoboo, Sure thing. I added the test case

oojacoboo: @michael-georgiadis sorry to be pedantic about this.  But can you please assert the value before the mutation and after, so we can be sure that the change is respected.  I don't actually recall if the previous mutations maintain state, or if each mutation within a test is stateless.

michael-georgiadis: @oojacoboo Made it more explicit 🫡

oojacoboo (approved):

tests/Integration/EndToEndTest.php
oojacoboo: @michael-georgiadis thanks

oojacoboo:
@michael-georgiadis Thanks for the PR and working through it with me!

michael-georgiadis:
@oojacoboo, no worries. Do we know when the release is coming up?

oojacoboo:
@michael-georgiadis I'm working on some other changes for 8.4, which this will be included.  Hope to have it completed soon.

